PR #15, opened 2024-01-10
Addressed issue #4

KianChilds:
## Justification
Allow for robot specific configuration without needing git to track the version specifically
This is a solution for Issue #4 

## Implementaion
Split the configuration into 2 sections. One is 'dynamic' which is updated each initialization. The other static that tracks things that should not change depend based on which robot it is on.

Dynamic will only contain sensor offsets and zeros
and static will contain all other configurations. Such as IDs, light patterns, setpoints, etc.

## Testing Done
None as of yet


zbrotherton:
In the future can we make the PR name match the feature?
Just so it's easier to determine team progress from the PR page

zbrotherton:
How does static config differ from a constants file?
Are we continuing constants in 2024

KianChilds:
There is no difference it's just inline with the naming convention we used last year

src/main/java/frc/robot/config/TunerConstants.java
azvanderpas, line 1: Since odds are good we will use Phoenix Tuner X to regenerate this file more than a few times this season, I'm not sure we should edit it directly much. 

Maybe having multiple TunerConstants files/classes with a common parent class/interface would work? Preferences could be used to select which constants class to instantiate.

KianChilds: The use of regenerating the file would be to rezero the pods. The current plan is add functionality to have our own rezero routine that would save to robot preferences.

KianChilds: Plan has updated, The process is now to copy the values from the generated file into shuffleboard

src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -0,0 +1,100 @@
...
+    if (initNT) {
+      initNT();
+    }
+  }
+
+  /** Creates network table entries even if not pre-existing */
+  public static void initNT() {
+    logInfo("Initlizing Dynamic Logs");
+    for (String key : allKeys) {
+      Preferences.initDouble(key, 0);
```
azvanderpas, line 80: Add a comment to clarify that initDouble() won't overwrite the value of any existing Preferences.

src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    if (initNT) {
+      initNT();
+    }
+  }
+
+  /** Creates network table entries even if not pre-existing */
+  public static void initNT() {
+    logInfo("Initlizing Dynamic Logs");
```
azvanderpas, line 78:
```suggestion
    logInfo("Initializing Dynamic Logs");
```

src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -0,0 +1,100 @@
...
+    frontLeftOffset_key, frontRightOffset_key, backRightOffset_key, backLeftOffset_key
+  };
+
+  public final double frontLeftOffset;
+  public final double frontRightOffset;
+  public final double backRightOffset;
+  public final double backLeftOffset;
+
+  public TunerConstants getTunerConstants() {
+    return new TunerConstants(frontLeftOffset, frontRightOffset, backLeftOffset, backRightOffset);
```
azvanderpas, line 33: Better to create a TunerConstants class member to return here. Otherwise we get a new instance each time this method is called.

src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -0,0 +1,100 @@
...
+      Preferences.initDouble(key, 0);
+    }
+  }
+
+  public static DynamicRobotConfig loadDynamicRobotConfig() {
+    if (dynRobotConfig != null) {
+      return dynRobotConfig;
+    }
+    return new DynamicRobotConfig();
+  }
```
azvanderpas, line 89: dynRobotConfig needs to be instantiated in this method for the singleton pattern I think you want. 

There's also a potential [thread safety concern](https://www.digitalocean.com/community/tutorials/thread-safety-in-java-singleton-classes), though the likelihood of us running into it is low.
Do you see a large benefit to implementing a singleton over a static class?

src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -0,0 +1,100 @@
...
+  // Contains every key that is used
+  // if one is missing it won't be correctly saved, or init
+  private static final String[] allKeys = {
+    frontLeftOffset_key, frontRightOffset_key, backRightOffset_key, backLeftOffset_key
+  };
+
+  public final double frontLeftOffset;
+  public final double frontRightOffset;
+  public final double backRightOffset;
+  public final double backLeftOffset;
```
azvanderpas, line 30: These should probably be private.

azvanderpas (changes requested):
How will the other generated values from Tuner X be stored? Just using StaticRobotConfig, or also with preferences?

If Preferences work in JUnit, I'd also recommend creating one or more unit tests to validate this works the way we think it will

src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -0,0 +1,100 @@
...
+      initNT = true;
+    }
+
+    if (Preferences.containsKey(backLeftOffset_key)) {
+      backLeftOffset = Preferences.getDouble(backLeftOffset_key, 0d);
+    } else {
+      raiseWarning("Back Left Pod Offset NOT FOUND!! using default");
+      backLeftOffset = 0;
+      initNT = true;
+    }
```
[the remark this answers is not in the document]
KianChilds: I considered using a dictionary but went against it for 3 reasons.
1. It would require all the preferences to have to the same type. (As in disallowing a String to be saved for a Limelight name)
2. The implementation for the configurability I wanted didn't seem cleaner. As in a function that takes in a error message, a key, and be able to raise a flag would be more difficult to read while sticking with common java types.
3. The conversion from a key to field would be difficult. then would also prohibit readability (because reflection code is confusing)

src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -0,0 +1,67 @@
...
+    return new TunerConstants(
+        ConfigVariables.frontLeftOffset,
+        ConfigVariables.frontRightOffset,
+        ConfigVariables.backLeftOffset,
+        ConfigVariables.backRightOffset);
+  }
+
+  public static DynamicRobotConfig loadDynamicRobotConfig() {
+    return new DynamicRobotConfig();
+  }
```
azvanderpas, line 62: This defeats the purpose of having a private constructor.

src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  public TunerConstants getTunerConstants() {
+    return new TunerConstants(
+        ConfigVariables.frontLeftOffset,
+        ConfigVariables.frontRightOffset,
+        ConfigVariables.backLeftOffset,
+        ConfigVariables.backRightOffset);
+  }
```
azvanderpas, line 58: Returning a new object any time this is called seems like a bad idea. Could wind up with two SwerveSubsystem instances.

src/main/java/frc/robot/config/TunerConstants.java
```diff
@@ -84,7 +85,6 @@ public class TunerConstants {
   private static final int kFrontLeftDriveMotorId = 1;
   private static final int kFrontLeftSteerMotorId = 2;
   private static final int kFrontLeftEncoderId = 1;
-  private static final double kFrontLeftEncoderOffset = -0.0361328125;
```
azvanderpas, line 87: Probably should go ahead and preload the DynamicRobotConfig with these values.

src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -0,0 +1,67 @@
...
+    // It then iterates through this list of references
+    for (Field variable : ConfigVariables.class.getFields()) {
+      // For each field, its name is stored as a string, so it can be used as a preference key
+      String key = variable.getName();
+      try {
+        // If the preference does not exist, or we are bypassing to competition robot defaults, log
+        // an alert and open network access for that preference
+        if (!Preferences.containsKey(key) || Robot.isCompetition) {
+          raiseWarning("Using competition robot default for " + key);
+          Preferences.initDouble(key, variable.getDouble(ConfigVariables.class));
```
azvanderpas, line 37: Consider how we will support other Preference data types.

azvanderpas (changes requested):

azvanderpas:
Some alternate structures for DynamicRobotConfig to consider (branched from the previous iteration):
[preference-map](https://github.com/frc6377/crescendo_2024/blob/preference-map/src/main/java/frc/robot/config/DynamicRobotConfig.java)
[functional-interface-preferences-map](https://github.com/frc6377/crescendo_2024/blob/functional-interface-preferences-map/src/main/java/frc/robot/config/DynamicRobotConfig.java)

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -54,11 +54,16 @@ public class RobotContainer {
   // driving in open loop
   private final SwerveRequest.SwerveDriveBrake brake = new SwerveRequest.SwerveDriveBrake();
   private final SwerveRequest.PointWheelsAt point = new SwerveRequest.PointWheelsAt();
-  private final Telemetry logger = new Telemetry(MaxSpeed);
+  private final Telemetry logger;
+
+  private final DynamicRobotConfig dynamicRobotConfig = DynamicRobotConfig.loadDynamicRobotConfig();
```
lohengrin332, line 59: Any reason not to initialize this within the constructor, like you are doing with the other variables you've changed in this class?

src/main/java/frc/robot/config/DynamicRobotConfig.java
zbrotherton: I think he intended to further develop it as protection against creating two classes? At the moment that doesn't work so I'll take it out

src/main/java/frc/robot/config/DynamicRobotConfig.java
zbrotherton: I could implement functionality for that, but as it stands can we foresee a use case for anything other than doubles?

azvanderpas: Booleans and Integers seem likely. They can be future iterations, but we won't want to have to re-write everything to get that support.

azvanderpas (approved):

azvanderpas (approved):

azvanderpas (approved):

PR #16, opened 2024-01-11
Shooter prototype code

ModBoyex:
This is code for shooter prototyping. This code should not be merged in to main, but is just here so that it can be reviewed


src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,80 @@
...
+
+
+  private Spark TLmotor1;
+  private Spark TLmotor2;
+  private Spark BRmotor1;
+  private Spark BRmotor2;
+  private Spark otherMotor;
+
+  public ShooterSubsystem() {
+    pid = new PIDController(1, 1, 1);
```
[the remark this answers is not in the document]
ModBoyex: Done

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,191 @@
...
+  private double feederVelo;
+
+  private CANSparkMax TLmotor1;
+  private CANSparkMax TLmotor2;
+  private CANSparkMax BRmotor1;
+  private CANSparkMax BRmotor2;
+  private CANSparkMax feederMotor;
+
+  public ShooterSubsystem() {
+    TLP = 1.0;
```
caseyjbrotherton, line 36: 1 is an aggressive P value for a sparkmax.  

Think you want to start with more like 1e-6

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,191 @@
...
+   * @return value of some boolean subsystem state, such as a digital sensor.
+   */
+  public boolean exampleCondition() {
+    // Query some boolean state, such as a digital sensor.
+    return false;
+  }
+
+  @Override
+  public void periodic() {
+    shooterVelo = SmartDashboard.getNumber("shooterVelo", shooterVelo);
```
caseyjbrotherton, line 126: Does shooterVelo or feederVelo adjust when the motors are running at all?  

Seems the only place shooterVelo is used is to setReference when the bumper is pressed.

And I don't think it will update every periodic with that.  Not sure what the intention is.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,191 @@
...
+        adjustBRP();
+    }
+    if(BRI != SmartDashboard.getNumber("BRShooterI", BRI)){
+        adjustBRI();
+    }
+    if(BRD != SmartDashboard.getNumber("BRShooterD", BRD)){
+        adjustBRD();
+    }
+
+    SmartDashboard.putNumber("shooterVelo", shooterVelo);
```
caseyjbrotherton, line 148: Is this to allow changing the shooterVelo through the triggers?  
Seems like that means depending on the timing that shooterVelo can be overwritten by either trigger or shuffleboard.  Not sure.

ModBoyex: Reshik made it so that it could be changed through the triggers, but I think it would make more sense just to make it changeable through Shuffleboard

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
ModBoyex: Done.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
ModBoyex: The intention is for you to be able to change the value in shuffle board

.gitignore
```diff
@@ -189,3 +189,4 @@ local.properties
 .vscode/settings.json
 *.wpilog
 
+src\main\java\frc\robot\BuildConstants.java
```
azvanderpas, line 192: Slashes are wrong direction. BuildConstants.java isn't getting ignored

azvanderpas:
Stale PR, closing

PR #19, opened 2024-01-11
Install Advantage Kit

ModBoyex:
Installing Advantage Kit for better data logging.


src/main/java/frc/robot/Robot.java
```diff
@@ -25,6 +32,31 @@ public class Robot extends TimedRobot {
...
+
+    if (isReal()) {
+      Logger.addDataReceiver(new WPILOGWriter()); // Log to a USB stick ("/U/logs")
+      Logger.addDataReceiver(new NT4Publisher()); // Publish data to NetworkTables
+      new PowerDistribution(1, ModuleType.kRev); // Enables power distribution logging
+    } else {
+      setUseTiming(false); // Run as fast as possible
+      String logPath =
+          LogFileUtil
+              .findReplayLog(); // Pull the replay log from AdvantageScope (or prompt the user)
```
azvanderpas, line 48: This doesn't seem to work when I try it in simulation.

caseyjbrotherton: `No log provided with the AKIT_LOG_PATH environment variable or through AdvantageScope. Enter path to file: Error at java.base/java.util.Scanner.nextLine(Scanner.java:1651): Unhandled exception: java.util.NoSuchElementException: No line found`


That's the error in the check.  Should we handle the exception and log it?  
Should we consider adding replay later?  
If I remember correctly, there are a couple of other things that need to happen for replay to work.

ModBoyex: I do think there are more steps to making replay work, but what feature does replay give us? is just reading the log file enough?

zbrotherton: I think I found the error - it seems to be because you don't have advantagescope open with a log file on your computer when simulating. It's not finding the file in tmpdir because it doesn't exist - advantagescope initializes that file when a log is opened.

zbrotherton: It will likely function, I know we don't have a log to try at the moment

zbrotherton: https://github.com/Mechanical-Advantage/AdvantageKit/blob/main/junction/core/src/org/littletonrobotics/junction/LogFileUtil.java

src/main/java/frc/robot/BuildConstants.java
azvanderpas, line 1: Need to delete this file from git tracking again

src/main/java/frc/robot/Robot.java
caseyjbrotherton: Can't resolve this.  Still doesn't work.  We need to either remove replay, or handle the exception.  

It prompts when you run on a laptop, but won't prompt when you run as part of COntinuous integration.  ( Tested again on my laptop today. )

src/main/java/frc/robot/BuildConstants.java
caseyjbrotherton: Deleted

lohengrin332 (dismissed):

lohengrin332 (approved):

PR #20, opened 2024-01-11
Feature/deltadebugboard

Josh-Westbrook:
## Justification
 only send new information to shuffleboard when not in a competition

## Implementaion
 combine necessary parts

## Testing Done
 build code


src/main/java/frc/robot/networktables/DeltaDebugBoard.java
```diff
@@ -0,0 +1,76 @@
...
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Robot;
+import java.util.HashMap;
+import java.util.Hashtable;
+import java.util.function.Consumer;
+
+public class DeltaDebugBoard<T> {
+  private static Hashtable<String, String> stringTable = new Hashtable<>();
+  private static Hashtable<String, Double> doubleTable = new Hashtable<>();
+  private static Hashtable<String, Boolean> booleanTable = new Hashtable<>();
```
zbrotherton, line 19: Do we use these hashtables?

src/main/java/frc/robot/networktables/DeltaDebugBoard.java
```diff
@@ -0,0 +1,76 @@
...
+  private static final HashMap<String, GenericEntry> entries = new HashMap<String, GenericEntry>();
+
+  private ShuffleboardTab networkTab;
+  private GenericEntry networkEntry;
+  private Integer localEntry;
+  private Consumer<T> localConsumer;
+  private final String name;
+  private T lastValue;
+
+  public DeltaDebugBoard(T defaultValue, String name, SubsystemBase subsystem) {
```
zbrotherton, line 31: Can we change this back to DebugLog?

DeltaDebugBoard isn't very readable

src/main/java/frc/robot/networktables/DeltaDebugBoard.java
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  public void log(T newValue) {
+    try {
+      if (!Robot.isCompetition) {
+        if (lastValue != newValue) {
+          networkEntry.setValue(newValue);
+        }
+        lastValue = newValue;
+      }
```
zbrotherton, line 70:
```suggestion
      if (!Robot.isCompetition && lastValue != newValue) {
        networkEntry.setValue(newValue);
        lastValue = newValue;
      }
```

.gitattributes
```diff
@@ -1 +1 @@
-*.* text eol=lf
\ No newline at end of file
+*.* text=auto eol=lf
```
zbrotherton, line 1: What does this do?

caseyjbrotherton: this is a cherry pick from 3b1a2709abc11f11d4822c9c3b936c7a0f7e1492

It prevents git from trying to convert LFs on non-text files.

azvanderpas: main already has this change though

zbrotherton: Got it, was just curious

src/main/java/frc/robot/networktables/DeltaDebugBoard.java
```diff
@@ -0,0 +1,76 @@
...
+  private GenericEntry networkEntry;
+  private Integer localEntry;
+  private Consumer<T> localConsumer;
+  private final String name;
+  private T lastValue;
+
+  public DeltaDebugBoard(T defaultValue, String name, SubsystemBase subsystem) {
+    this.name = name;
+    this.lastValue = defaultValue;
+    if (defaultValue instanceof Double) {
```
azvanderpas, line 34: Would be useful to catch and report errors on attempts to use an unsupported data type.

.gitattributes
caseyjbrotherton: I cherry picked it to all active branches to try and prevent gradle from being broken.

src/main/java/frc/robot/networktables/DeltaDebugBoard.java
```diff
@@ -0,0 +1,76 @@
...
+        networkTab = Shuffleboard.getTab(subsystem.getName());
+
+        if (!entries.containsKey(name)) {
+          networkEntry = networkTab.add(name, defaultValue).getEntry();
+          entries.put(name, networkEntry);
+        } else {
+          networkEntry = entries.get(name);
+          DriverStation.reportWarning(
+              "Duplicate ShuffleboardEntry on " + networkTab.getTitle() + " tab: " + name, false);
+        }
```
azvanderpas, line 57: Is `entries` only there to detect duplicate names? If so, I'm not sure we need to keep track of the GenericEntry values. Consider using [HashSet](https://docs.oracle.com/javase/8/docs/api/java/util/HashSet.html) instead of HashMap.

src/main/java/frc/robot/networktables/DeltaDebugBoard.java
azvanderpas: I think we should append -Entry to the class name, whichever we use. Since this initializes an single entry, and not an entire log/board.

src/main/java/frc/robot/networktables/DeltaDebugBoard.java
```diff
@@ -0,0 +1,76 @@
...
+      if (!Robot.isCompetition) {
+        networkTab = Shuffleboard.getTab(subsystem.getName());
+
+        if (!entries.containsKey(name)) {
+          networkEntry = networkTab.add(name, defaultValue).getEntry();
+          entries.put(name, networkEntry);
+        } else {
+          networkEntry = entries.get(name);
+          DriverStation.reportWarning(
+              "Duplicate ShuffleboardEntry on " + networkTab.getTitle() + " tab: " + name, false);
```
azvanderpas, line 56: This looks like a case to cover in a Unit test.  #24 
Doesn't need to be added for this PR

src/main/java/frc/robot/networktables/DeltaDebugBoard.java
zbrotherton: agreed

zbrotherton: Settle on DebugEntry?

As much as I love DeltaDebugBoardEntry

src/main/java/frc/robot/networktables/DeltaDebugBoard.java
zbrotherton: It's intended to allow multiple references to the same log in different subsystems - a duplicate will just be directed to the existing entry so it can still function

zbrotherton (approved):

caseyjbrotherton (approved):

PR #37, opened 2024-01-13
Create spotlessApply pre-commit

### LINKED ISSUE #33: Add spotlessApply to Pre-Commit

RobbyV2:
It's possible to make it so that on build spotlessApply is also executed.

## Justification
Fixes #33 

## Implementaion
Adds a gradle build task and a script that is copied to the local .git/hooks folder.

## Testing Done
Checked that pre-commit is in the .git/hooks folder and it runs `spotlessApply` before a commit. 

## Notes
It's possible to make it so that on build `spotlessApply` is also executed.
Although, that would also require some more work to try and figure out how to get this task running before any other one. Since tasks such as `spotlessCheck` currently run before `spotlessApply` has a chance to run, which makes the build fail.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -269,3 +269,16 @@ deploy.targets.roborio.artifacts.frcStaticFileDeploy.dependsOn(writeComputerName
 //             "date.txt"
 //             ).text = outstream
 // }
+
+tasks.register("addPreCommitGitHookOnBuild") {
+    exec {
+        workingDir projectDir
+        if (System.getProperty('os.name').toLowerCase(Locale.ROOT).contains('windows')) {
+            commandLine 'cmd', '/c', 'copy', 'scripts\\pre-commit', '.git\\hooks\\pre-commit'
+        } else {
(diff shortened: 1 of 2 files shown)


lohengrin332 (approved):
This looks good to me, but I'd like to see whether anyone else has any concerns.

caseyjbrotherton (changes requested):
Spotless DOesn't like BuildConstants.java that gets created from advantagekit.

Need to do something about that before committing this

https://stackoverflow.com/questions/41180662/ignore-generated-resources-from-spotless

Maybe try this?

caseyjbrotherton:
> Spotless DOesn't like BuildConstants.java that gets created from advantagekit.
> 
> Need to do something about that before committing this
> 
> https://stackoverflow.com/questions/41180662/ignore-generated-resources-from-spotless
> 
> Maybe try this?

@ModBoyex added this to the advantagekit pull request.  Rest of it looks good to me.

caseyjbrotherton (approved):

PR #38, opened 2024-01-13
Input mixing draft

Shazamboo:
## Justification
 Need to fine tune input for driving

## Implementaion
 made a new OI class to manage buttons and calculate the control curve for swerve movement
 could be useless except for the control curve part

## Testing Done
 none yet


src/main/java/frc/robot/OI.java
```diff
@@ -0,0 +1,232 @@
...
+
+    for (Operator.Button button : Operator.Button.values()) {
+      operatorButtonsLayout.add(
+          String.valueOf(button.getButtonID() + Operator.Button.values().length),
+          "Button " + button.toString() + ": " + button.getButtonAction());
+    }
+  }
+
+  public static class ControlCurve {
+    private double ySaturation; // Maximum output, in percentage of possible output
```
lohengrin332, line 197: It looks like this value is set when the class is instantiated, then never updated again. If that is correct, you should declare it `final`. The same goes for your function parameters below and elsewhere. `final` not only causes a variable to be unchangeable, it also _requires_ that it be set, issuing a compilation error when it's not, and also treats the variables differently during the compilation process. If you're interested, [here](https://stackoverflow.com/questions/27369584/what-does-the-jvm-do-with-a-final-variable-in-a-class) is a good SO Q&A about this difference.

src/main/java/frc/robot/OI.java
```diff
@@ -0,0 +1,232 @@
+package frc.robot;
+
+import edu.wpi.first.wpilibj.Joystick;
+import edu.wpi.first.wpilibj.shuffleboard.BuiltInLayouts;
+import edu.wpi.first.wpilibj.shuffleboard.Shuffleboard;
+import edu.wpi.first.wpilibj.shuffleboard.ShuffleboardLayout;
+import edu.wpi.first.wpilibj2.command.button.JoystickButton;
+import java.util.Map;
+
+public class OI {
```
lohengrin332, line 10: Is there a good reason not to call this class `OperatorInterface` instead of `OI`?

PR #42, opened 2024-01-14
Log Pose2D & Rotation

Asher123456789:
We added Pose2D and its rotation to SmartDashboard.

### DIFF
--- a/src/main/java/frc/robot/Telemetry.java
+++ b/src/main/java/frc/robot/Telemetry.java
@@ -3,12 +3,7 @@ package frc.robot;
 import com.ctre.phoenix6.Utils;
 import com.ctre.phoenix6.mechanisms.swerve.SwerveDrivetrain.SwerveDriveState;
 import edu.wpi.first.math.geometry.Pose2d;
-import edu.wpi.first.math.geometry.Rotation2d;
 import edu.wpi.first.math.geometry.Translation2d;
-import edu.wpi.first.math.kinematics.SwerveDriveKinematics;
-import edu.wpi.first.math.kinematics.SwerveDriveWheelPositions;
-import edu.wpi.first.math.kinematics.SwerveModuleState;
-import edu.wpi.first.math.util.Units;
 import edu.wpi.first.networktables.DoubleArrayPublisher;
 import edu.wpi.first.networktables.DoublePublisher;
 import edu.wpi.first.networktables.NetworkTable;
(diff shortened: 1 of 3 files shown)


azvanderpas (changes requested):

src/main/java/frc/robot/Telemetry.java
```diff
@@ -25,11 +26,14 @@ public class Telemetry {
    */
   public Telemetry(double maxSpeed) {
     MaxSpeed = maxSpeed;
```
lohengrin332, line 28: I know this is pre-exiating code, but idiomatic Java usually uses `lowerCaseFirst` naming for variables, and `UpperCaseFirst` for classes. Please update this variable accordingly, and then you can differentiate it from the parameter of the same name using `this.maxSpeed`. E.g.:

```suggestion
    this.maxSpeed = maxSpeed;
```

src/main/java/frc/robot/Telemetry.java
```diff
@@ -25,11 +26,14 @@ public class Telemetry {
...
+    SmartDashboard.putData(field);
   }
 
   /* What to publish over networktables for telemetry */
   NetworkTableInstance inst = NetworkTableInstance.getDefault();
 
+  Field2d field = new Field2d();
```
lohengrin332, line 35: Please use `final` for variables whenever possible. See [this Q&A from StackOverflow](https://stackoverflow.com/questions/27369584/what-does-the-jvm-do-with-a-final-variable-in-a-class) for one reason why.

```suggestion
  final Field2d field = new Field2d();
```

src/main/java/frc/robot/Telemetry.java
```diff
@@ -25,11 +26,14 @@ public class Telemetry {
...
+
   /* Robot pose for field positioning */
   NetworkTable table = inst.getTable("Pose");
   DoubleArrayPublisher fieldPub = table.getDoubleArrayTopic("robotPose").publish();
```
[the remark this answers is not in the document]
ModBoyex: but it doesn not allow you to drive the robot in simulation. Maybe there a better way to do it, but what Asher did allowed you to drive it.

src/main/java/frc/robot/Telemetry.java
ModBoyex: This gives me an error when changing it to this. I leave that for another pull request.

src/main/java/frc/robot/Telemetry.java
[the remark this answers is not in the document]
ModBoyex: Never mind, found were it logs it in networktables. Asher's code is now obsolete.

PR #45, opened 2024-01-17
(Merge of InputMixing and RumbleCompatibility) Add OI Class, move swerve state management to subsystem

zbrotherton:
## Justification

Messy PR but this adds inputMixing and moves towards a cleaner codebase

## Testing Done

Not tested yet


src/main/java/frc/robot/CommandSwerveDrivetrain.java
```diff
@@ -17,6 +21,13 @@
  */
 public class CommandSwerveDrivetrain extends SwerveDrivetrain implements Subsystem {
   private static final double kSimLoopPeriod = 0.005; // 5 ms
+  private static final double maxSpeed = Units.feetToMeters(18.2); // Desired top speed
+  private static final double maxAngularRate = Math.PI; // Max angular velocity
+  private final Telemetry telemetry = new Telemetry(maxSpeed); //TODO unused?
```
KianChilds, line 26: Comment is inaccurate

src/main/java/frc/robot/OI.java
```diff
@@ -0,0 +1,235 @@
...
+        this.buttonAction = name;
+      }
+    };
+    
+    private static final Joystick joystick = new Joystick(driverJoystickPort);
+    private static final XboxController rumbleController = new XboxController(driverJoystickPort);
+    
+    private static final Button orientationButton = Button.Start; // Toggle swerve orientation
+    private static final Button outtakeButton = Button.RB; // Run outtake
+    private static final Button intakeButton = Button.LB; // Run intake
```
KianChilds, line 64: This is a button change from main
In main intake is left trigger

src/main/java/frc/robot/OI.java
KianChilds, line 1: Does this class support trigger input? such as in main how left trigger is used to activate the intake

src/main/java/frc/robot/CommandSwerveDrivetrain.java
```diff
@@ -55,6 +68,42 @@ private void startSimThread() {
     m_simNotifier.startPeriodic(kSimLoopPeriod);
   }
 
+  public SwerveRequest drive(double x, double y, double r){
```
KianChilds, line 71: The method is in accurate. This function isn't telling the drive train to move, it is getting the request to submit to tell the drive train to move. something along the the lines of `getDriveRequest(...)` would be a better name

Also comment stating the expected input range would be useful, and changing r to a different name.
r could be thought as rotation but that isn't what it represents. With it representing angular velocity it would be closer to alpha then r.

KianChilds (changes requested):

src/main/java/frc/robot/OI.java
```diff
@@ -0,0 +1,235 @@
...
+
+    public ControlCurve(double ySaturation, double yIntercept, double curvature, double deadzone) {
+      this.ySaturation = ySaturation;
+      this.yIntercept = yIntercept;
+      this.curvature = curvature;
+      this.deadzone = deadzone;
+    }
+
+    public double calculate(double input) {
+      /* https://www.desmos.com/calculator/w6ovblmmqj
```
tylerhedge, line 213: consider changing to a continuous function like in this desmos
https://www.desmos.com/calculator/ilq3nd4rsg

src/main/java/frc/robot/OI.java
```diff
@@ -0,0 +1,235 @@
...
+
+public class OI {
+  // Operator Interface (OI) class containing all control information
+
+  private static final int driverJoystickPort = 0;
+  private static final int operatorJoystickPort = 1;
+
+  public static final class Driver {
+
+    private static enum Button {
```
tylerhedge, line 21: is most of this class needed given the existence of the CommandXboxController class?

azvanderpas (changes requested):

src/main/java/frc/robot/OI.java
```diff
@@ -0,0 +1,235 @@
...
+      }
+
+      private void setButtonAction(String name) {
+        this.buttonAction = name;
+      }
+    };
+
+    private static final Joystick joystick = new Joystick(OI.operatorJoystickPort);
+    private static final Button kExampleButton =
+        Button.LB; // Toggles intake mode between cone and cube
```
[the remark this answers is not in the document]
zbrotherton: Not a consequential fix but fair enough, taking out the example button entirely

src/main/java/frc/robot/OI.java
zbrotherton: Added trigger support

src/main/java/frc/robot/OI.java
zbrotherton: This class was created to allow controls to be displayed on a shuffleboard tab - so the driver is never confused when programming changes them

src/main/java/frc/robot/OI.java
```diff
@@ -0,0 +1,235 @@
...
+      return new JoystickButton(joystick, orientationButton.getButtonID());
+    }
+
+    public static JoystickButton getIntakeButton() {
+      intakeButton.setButtonAction("Intake");
+      return new JoystickButton(joystick, intakeButton.getButtonID());
+    }
+
+    public static JoystickButton getOuttakeButton() {
+      outtakeButton.setButtonAction("Outtake");
```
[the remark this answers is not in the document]
zbrotherton: sure

src/main/java/frc/robot/OI.java
zbrotherton: I could rewrite as a wrapper on the xboxcontroller class? This is pretty much exactly cut from last year

src/main/java/frc/robot/OI.java
zbrotherton: Input immediately out of deadzone should be just enough to overcome static friction in my experience.

src/main/java/frc/robot/OI.java
caseyjbrotherton: triggerDetent added

caseyjbrotherton (dismissed):

src/main/java/frc/robot/OI.java
caseyjbrotherton: Thanks Tyler.  Think this function is fine for now, and we can change to continuous after testing if needed.  Resolving

src/main/java/frc/robot/OI.java
tylerhedge: if there is value-added compared to the CommandXboxController, and the code is already written I dont have any complaints. I just want to make sure we are using this code for a reason

src/main/java/frc/robot/CommandSwerveDrivetrain.java
```diff
@@ -55,6 +64,55 @@ private void startSimThread() {
...
+    return new SwerveRequest.SwerveDriveBrake();
+  }
+
+  public SwerveRequest getAlignRequest(Rotation2d rotation) {
+    return new SwerveRequest.FieldCentricFacingAngle().withTargetDirection(rotation);
+  }
+
+  public void toggleOrientation() {
+    isFieldOriented = !isFieldOriented;
+    if (!isFieldOriented) {
```
tylerhedge, line 103: do you need an 
```
else:
    setAlignment(wayToGetAlignment);
```

azvanderpas: Worth adding a comment to show intention (robot oriented drive means we don't want to/can't hold a field oriented heading)

zbrotherton:
Thoughts on splitting into two files - OI and OI Utilities?
So that OI is more concise

src/main/java/frc/robot/OI.java
```diff
@@ -18,225 +19,191 @@ public class OI {
...
+    private ControlCurve getCurve() {
+      return curve;
+    }
 
-      private double getTriggerDetent() {
-        return this.triggerDetent;
-      }
+    private double getDetent() {
+      return detent;
+    }
```
azvanderpas, line 160: Spelling? Or is this a word I don't know?

zbrotherton: Maybe threshold is a better word, I'm just used to it being called a detent in triggers

caseyjbrotherton: agreed.  Threshold is better.

src/main/java/frc/robot/OI.java
```diff
@@ -0,0 +1,242 @@
...
+
+  public static final class Operator {
+    private static final XboxController controller = new XboxController(operatorJoystickPort);
+  }
+
+  // --- OI UTILITIES -- //
+
+  public static Supplier<Double> getAxisSupplier(Control axis) {
+    if (axis.getType() != Control.ControlType.AXIS) {
+      DriverStation.reportWarning(axis.getAction() + " is not an axis", true);
```
caseyjbrotherton, line 59: Should be error?

src/main/java/frc/robot/OI.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private ControlType getType() {
+      return type;
+    }
+
+    private void putControl() {
+      if (controller.getPort() == driverJoystickPort) {
+        driverControlsLayout.add(
+            "Driver" + String.valueOf(getId()),
```
azvanderpas, line 169: This doesn't work because you Control id's are not unique. e.g. there is an axis of id 1 (Left Y axis) and a button of id 1 (A button).

zbrotherton:
This should probably be merged ASAP - PR is really bloated, but it is fully driver approved and tested

caseyjbrotherton (approved):
Has been tested on robot

PR #54, opened 2024-01-18
Fix Duplicate NT Entry & JUnit Testing

### LINKED ISSUE #35: Fix TestDebugEntry JUnit tests
I'd like to revisit these tests. I have some concerns about them

- DebugEntry is the class under test, so I don't think it makes sense to be using it with MockedConstruction. 
- There is a 'gotcha' here with the JUnit test environment. Each test gets run one after the other without tearing down anything in between. So all the new DebugEntry()'s that get made in one test still exist when the next test is run. For this reason, either each test should use unique keys (i.e. not all "test"), or a new method to tear down the DebugEntry HashMap must be made with the @AfterEach annotation.
- Another test to validate correct functionality should be added. Specifically that two DebugEntries with the same name can be made without error/warning as long as they are for different subsystems.

https://github.com/frc6377/crescendo_2024/blob/a9db5335104cfb21d7cbd5fd5f9b985e62c30fb7/src/test/java/TestDebugEntry.java#L39-L81

### LINKED ISSUE #50: Duplicate DebugEntries should cause an error
https://github.com/frc6377/crescendo_2024/blob/a2c21b601ebd0fd81ddc30770137723aa9127814/src/main/java/frc/robot/networktables/DebugEntry.java#L49-L55

Duplicate names currently cause a warning, but then assigns the NT entry to the existing one. This doesn't seem like a good idea, since now the NT entry has multiple sources, and the default value can get overridden with the current code

### LINKED ISSUE #53: remove exampleSubsystem
We should clean up and remove exampleSubsystem

RobbyV2:
Fixes #35 (adds teardown of tests)
Fixes #50 (raises exception/error, stops it from resetting the duplicate NT entry)
Fixes #53 (removes example subsystem and how robot container relies on it)


src/main/java/frc/robot/networktables/DebugEntry.java
```diff
@@ -47,9 +47,11 @@ public DebugEntry(T defaultValue, String name, SubsystemBase subsystem) {
           networkEntry = networkTab.add(name, defaultValue).getEntry();
           entries.put(name, networkEntry);
         } else {
-          networkEntry = entries.get(name);
-          DriverStation.reportWarning(
-              "Duplicate ShuffleboardEntry on " + networkTab.getTitle() + " tab: " + name, false);
+          // networkEntry = entries.get(name);
+          String errorMessage =
+              "Duplicate ShuffleboardEntry on " + networkTab.getTitle() + " tab: " + name;
+          DriverStation.reportError(errorMessage, false);
```
caseyjbrotherton, line 53: REmoving this message doesn't cause a test error.

I think that the throw is skipping part of your test.  

Make sure to change the pieces of the code to cause a failure and observe that the test fails when it should

caseyjbrotherton (changes requested):

caseyjbrotherton (approved):
Looks good, tested locally

PR #55, opened 2024-01-18
Turret odometry

caseyjbrotherton:


src/main/java/frc/robot/commands/Turret/TurretCommand.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {}
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    turretSubsystem.setTurretVelo(1);
```
caseyjbrotherton, line 35: Can we make this into a command factory in the turret subsystem?  There is a lot of template code that we could remove for a more readable solution.  

A Command Factory means that a method in the turret subsystem will return a command that can be used to assign to a trigger or button

src/main/java/frc/robot/commands/Turret/TurretMagicCommand.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {}
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    turretSubsystem.magicMethod(limelightRotation);
```
caseyjbrotherton, line 39: Can we rename this to be more descriptive of the magic?  :)

caseyjbrotherton: Is this tracking the updated heading from the limelight?  

The use case is aligning the turret to a goal while the drivebase is moving.  And I could be wrong, but this seems to just get the rotation at the beginning of the command.  Am I missing something with the BiConsumer piece of it?

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,93 @@
...
+    return turretMotor.getEncoder().getPosition();
+  }
+
+  @Override
+  public void periodic() {
+    
+  }
+
+  public double turretFromOdometry(Pose2d robotPos){
+    return Math.atan(robotPos.getY()/robotPos.getX()) + robotPos.getRotation().getDegrees();
```
caseyjbrotherton, line 81: Not sure what the intention of this is?  Can you please describe it further

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,93 @@
...
+
+    turretMotor.enableSoftLimit(CANSparkMax.SoftLimitDirection.kForward, true);
+    turretMotor.enableSoftLimit(CANSparkMax.SoftLimitDirection.kReverse, true);
+
+    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kForward, 1);
+    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kReverse, -1);
+
+    // initialze PID controller and encoder objects
+    turretPIDController = turretMotor.getPIDController();
+    m_encoder = turretMotor.getEncoder();
```
azvanderpas, line 46: This is the relative encoder embedded in the motor/sparkmax controller. We need to use the absolute encoder (CAN Coder) the turret is on to control the position.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,93 @@
...
+    turretPIDController.setI(Constants.TurretConstants.TURRET_KI);
+    turretPIDController.setD(Constants.TurretConstants.TURRET_KD);
+    turretPIDController.setIZone(Constants.TurretConstants.TURRET_KIZ);
+    turretPIDController.setFF(Constants.TurretConstants.TURRET_KFF);
+    turretPIDController.setOutputRange(Constants.TurretConstants.TURRET_KMINOUTPUT, Constants.TurretConstants.TURRET_KMAXOUTPUT);
+
+    turretPIDController.setSmartMotionMaxVelocity(Constants.TurretConstants.TURRET_MAXVEL, Constants.TurretConstants.TURRET_SMARTMOTION_SLOT);
+    turretPIDController.setSmartMotionMinOutputVelocity(Constants.TurretConstants.TURRET_MINVEL, Constants.TurretConstants.TURRET_SMARTMOTION_SLOT);
+    turretPIDController.setSmartMotionMaxAccel(Constants.TurretConstants.TURRET_MAXACC, Constants.TurretConstants.TURRET_SMARTMOTION_SLOT);
+    turretPIDController.setSmartMotionAllowedClosedLoopError(Constants.TurretConstants.TURRET_ALLOWEDERR, Constants.TurretConstants.TURRET_SMARTMOTION_SLOT);
```
azvanderpas, line 59: Since the CANCoder isn't supported by the SparkPIDController, we'll need to use PIDController from WPILib and can't use these smart motion features.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) FIRST and other WPILib contributors.
+// Open Source Software; you can modify and/or share it under the terms of
+// the WPILib BSD license file in the root directory of this project.
+
+package frc.robot.subsystems;
+
+import com.revrobotics.CANSparkLowLevel.MotorType;
+
+import javax.swing.text.Position;
```
azvanderpas, line 9: Unused import

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,93 @@
...
+
+import com.revrobotics.CANSparkMax;
+import com.revrobotics.RelativeEncoder;
+import com.revrobotics.SparkPIDController;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Rotation2d;
+import edu.wpi.first.wpilibj2.command.Command;
```
azvanderpas, line 17: Unused import

src/main/java/frc/robot/CommandSwerveDrivetrain.java
azvanderpas, line 1: Telemetry.java updates the robot pose already. I don't think the edits to this file are necessary

src/main/java/frc/robot/subsystems/TurretSubsystem.java
Shazamboo: You will have to ask rishik because I am lost as to what this does.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,93 @@
...
+    turretPIDController.setReference(velo, CANSparkMax.ControlType.kVelocity);
+  }
+
+  public double getTurretPos(){
+    return turretMotor.getEncoder().getPosition();
+  }
+
+  @Override
+  public void periodic() {
+
```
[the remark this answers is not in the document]
Shazamboo: not sure how debug entry works so I made shuffleboard entrys

src/main/java/frc/robot/commands/Turret/TurretCommand.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  // Called once the command ends or is interrupted.
+  @Override
+  public void end(final boolean interrupted) {
+    turretSubsystem.setTurretVelo(0.0);
```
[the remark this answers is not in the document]
Shazamboo: done

azvanderpas (changes requested):
Also might want to look at adding [feedforward](https://docs.wpilib.org/en/stable/docs/software/advanced-controls/controllers/combining-feedforward-feedback.html) to the PID control, as well as [profiled](https://docs.wpilib.org/en/stable/docs/software/advanced-controls/controllers/profiled-pidcontroller.html) motion

src/main/java/frc/robot/subsystems/TurretSubsystem.java
rishik4: It calculates how much the turret has to move given the position of the robot by using trig to calculate the X and Y offset from the target and then the offset angle of the robot and add that together to calculate the amount the turret has to move

src/main/java/frc/robot/CommandSwerveDrivetrain.java
```diff
@@ -36,6 +44,16 @@ public CommandSwerveDrivetrain(
     if (Utils.isSimulation()) {
       startSimThread();
     }
+
+    SmartDashboard.putData("Field", field);
+  }
+
+  public BiConsumer<Pose2d, Double> getVisionMeasurementConsumer() {
+    return (t, u) -> addVisionMeasurement(t, u);
+  }
```
azvanderpas, line 53: Keep this, remove the other changes to this file

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,109 @@
...
+    turretPIDController.setD(Constants.TurretConstants.TURRET_KD);
+    turretPIDController.setIZone(Constants.TurretConstants.TURRET_KIZ);
+  }
+
+  public void stopTurret() {
+    turretMotor.stopMotor();
+  }
+
+  public void setTurretPos(double setpoint) {
+    turretMotor.set(turretPIDController.calculate(turretPosition, setpoint));
```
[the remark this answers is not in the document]
azvanderpas: Additionally, we will want to always be sending updated values to the motor so that it holds it's position even when we aren't actively commanding it. So it will react to a defender collision causing it to shift, for instance. Consider either setting up a default command that holds position or always calling turretMotor.set() in periodic.

azvanderpas (changes requested):

src/main/java/frc/robot/CommandSwerveDrivetrain.java
Malachi-lewis: Why? Does it conflict with changes made on main, or is there some other reason?

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,109 @@
...
+        (m_encoder.getVelocity().getValueAsDouble())
+            * 60; // changing from rotations per second to rotations per minute or rpm
+    turretPositionEntry.log(turretPosition);
+    turretVelocityEntry.log(turretVelocity);
+  }
+
+  public double turretFromOdometry(
+      Pose2d robotPos) { // Doesn't work right now but will be used later
+    return Math.atan(robotPos.getY() / robotPos.getX()) + robotPos.getRotation().getDegrees();
+  }
```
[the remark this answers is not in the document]
Malachi-lewis: No? We need the X and Y because we're having the turret track a target just from robot position without any AprilTags. This wouldn't do anything without the X and Y.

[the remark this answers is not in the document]
azvanderpas: Currently this tracks the line from the field origin through the robot's position, which does not lead to a static position/target as the robot moves. 

turretFromOdometry() is also not a descriptive method name for the intent described.

azvanderpas (changes requested):
Recommend setting up simulation to catch issues before trying it on hardware

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,154 @@
...
+  public void TurretOdomCommand(Pose2d robotPos, Pose2d targetPos) {
+    turretPIDController.reset();
+    setTurretPos(getTurretRotationFromOdometry(robotPos, targetPos));
+  }
+
+  public Command buildTurretCommand(boolean limelightVisible, Pose2d robotPos, Pose2d targetPos) {
+    return limelightVisible
+        ? run(this::LockTurret)
+        : runEnd(() -> TurretOdomCommand(robotPos, targetPos), this::stopTurret);
+  }
```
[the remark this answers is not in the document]
Malachi-lewis: fixed

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,154 @@
...
+    zeroTurretEncoder();
+    turretPIDController.setIZone(Constants.TurretConstants.KIZ);
+  }
+
+  public void stopTurret() {
+    turretMotor.stopMotor();
+  }
+
+  public Command stowTurret() {
+    return new InstantCommand(() -> setTurretPos(Math.toRadians(110)));
```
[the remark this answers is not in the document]
Malachi-lewis: fixed

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  @Override
+  public void simulationPeriodic() {
+    updateTurretPosition();
+    turretVelocity =
+        (m_encoder.getVelocity().getValueAsDouble())
+            * 60; // changing from rotations per second to rotations per minute or rpm
+    turretPositionEntry.log(turretPosition);
+    turretVelocityEntry.log(turretVelocity);
```
azvanderpas, line 194: No need to duplicate periodic(), it runs in simulation already.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,210 @@
...
+    for (var i = 0; i < Robot.defaultPeriodSecs; i += CANSparkMaxSim.kPeriod) {
+      turretSim.setInput(turretMotor.get() * RobotController.getBatteryVoltage());
+      turretSim.update(CANSparkMaxSim.kPeriod);
+      turretMotor.update(
+          turretSim.getVelocityRadPerSec() * Constants.TurretConstants.CONVERSION_FACTOR);
+    }
+    turretAngleSim.setAngle(Math.toDegrees(turretSim.getAngleRads()));
+    SmartDashboard.putNumber("Turret Angle", Math.toDegrees(turretSim.getAngleRads()));
+  }
```
azvanderpas, line 203: Since we are using the PIDController class and not the SparkPIDController, remove the for loop.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,210 @@
...
+
+  private CANSparkMaxSim turretMotor;
+  private SingleJointedArmSim turretSim;
+  private Mechanism2d turretMech;
+  private MechanismRoot2d root;
+  private MechanismLigament2d turretAngleSim;
+  private ShuffleboardTab turretTab = Shuffleboard.getTab("Turret Tab");
+
+  private PIDController turretPIDController;
+  private CANcoder m_encoder;
```
azvanderpas, line 47: Look at #101 for an example of how to setup the CANcoder position as a SimDouble

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,210 @@
...
+import frc.robot.Robot;
+import frc.robot.stateManagement.RobotStateManager;
+import frc.robot.utilities.DebugEntry;
+import frc.robot.utilities.LimelightHelpers;
+import frc.robot.utilities.TunableNumber;
+import java.util.function.Consumer;
+
+public class TurretSubsystem extends SubsystemBase {
+
+  private CANSparkMaxSim turretMotor;
```
azvanderpas, line 39: CANSparkMaxSim shouldn't be needed here since we aren't using the SparkMAX PID Controllers

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,210 @@
...
+  private CANcoder m_encoder;
+  private double turretPosition;
+  private double turretVelocity;
+  private Consumer<Double> testPosition;
+  private DebugEntry<Double> turretPositionEntry =
+      new DebugEntry<Double>(turretPosition, "Position", this);
+  private DebugEntry<Double> turretVelocityEntry =
+      new DebugEntry<Double>(turretVelocity, "Velocity", this);
+  private TunableNumber tunableTestPosition =
+      new TunableNumber("Turret Test Position", 30, testPosition);
```
azvanderpas, line 56: TunableNumber seems to not work well right now. Suggest just using a Shuffleboard/Smartdashboard entry for now.

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,210 @@
...
+    turretPIDController.setIZone(Constants.TurretConstants.KIZ);
+  }
+
+  public void stopTurret() {
+    turretMotor.stopMotor();
+  }
+
+  public Command stowTurret() {
+    return new InstantCommand(() -> setTurretPos(Math.toRadians(0)));
+  }
```
azvanderpas, line 111:
```suggestion
  public Command stowTurret() {
    return new InstantCommand(() -> setTurretPos(Math.toRadians(0))).withName("StowTurretCommand");
  }
```

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,210 @@
...
+    return turretPosition; // returns the absolute encoder position in radians
+  }
+
+  public double getTurretVel() {
+    return turretVelocity;
+  }
+
+  public Command testTurretCommand() {
+    return runEnd(() -> setTurretPos(Math.toRadians(tunableTestPosition.get())), this::stopTurret);
+  }
```
azvanderpas, line 161:
```suggestion
  public Command testTurretCommand() {
    return runEnd(() -> setTurretPos(Math.toRadians(tunableTestPosition.get())), this::stopTurret).withName("TestTurretCommand");
  }
```

src/main/deploy/branch.txt
azvanderpas, line 1: Remove txt files from git tracking

src/main/java/frc/robot/Robot.java
azvanderpas, line 1: Revert this file

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+    return turretPosition; // returns the absolute encoder position in radians
+  }
+
+  public double getTurretVel() {
+    return turretVelocity;
+  }
+
+  public Command testTurretCommand() {
+    return runEnd(() -> setTurretPos(Math.toRadians(-60)), this::stopTurret);
+  }
```
azvanderpas, line 171: This would be useful to set to a specific angle as we need. Maybe modify to make it take a angle parameter (i.e. `testTurretCommand(double degrees)`)

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+
+  public void AimTurret() {
```
azvanderpas, line 173: Nitpick: method name should start with lowercase letter. Same with other AimTurretOdometry method

Also make this method private to encourage using the Command factories

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+    double limelightTX = LimelightHelpers.getTX("limelight");
+    if (limelightTX != 0
+        && LimelightHelpers.getFiducialID("limelight")
+            == ((robotStateManager.getAllianceColor() == AllianceColor.BLUE)
+                ? Constants.TurretConstants.SPEAKER_TAG_ID_BLUE
+                : Constants.TurretConstants.SPEAKER_TAG_ID_RED)) {
```
azvanderpas, line 179: If there are more than one April Tags in view, this might not always work, right? getFiducialID() gets the "primary in-view AprilTag"

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+
+      // Y & Tilting
+      double limelightTY = LimelightHelpers.getTY("limelight");
+      double distanceToTag = tyToDistanceFromTag(limelightTY);
+      tagDistanceEntry.log(distanceToTag);
+      // TODO: Add vertical tilt and use distance for it
+
+      if (Math.abs(Math.toRadians(limelightTX) + turretPosition)
+          > Math.toRadians(Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES)) {
+        // TODO: Make turret rotate the drivebase if necessary and driver thinks it's a good idea
```
azvanderpas, line 191: This is probably better done as a composite Command setup somewhere else, not in this subsystem

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+
+  public void holdPosition() {
+    setTurretPos(turretPosition);
+  }
+
+  public void zeroTurretEncoder() {
+    m_encoder.setPosition(0.0);
+  }
+
+  public void updateTurretPosition() {
```
azvanderpas, line 148: Lets make this private

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/TurretSubsystem.java
Malachi-lewis: We can have the limelight filter out all AprilTags except the one on each speaker we care about. Since we'll never see both of them, this won't be a problem.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+
+  private PIDController turretPIDController;
+  private CANcoder m_encoder;
+  private double turretPosition;
+  private double turretVelocity;
+  private Consumer<Double> testPosition;
+  private ShuffleboardTab configTab = Shuffleboard.getTab("Config");
+  private GenericEntry kP = configTab.add("Turret KP", Constants.TurretConstants.KP).getEntry();
+  private GenericEntry kI = configTab.add("Turret KI", Constants.TurretConstants.KI).getEntry();
+  private GenericEntry kD = configTab.add("Turret KD", Constants.TurretConstants.KD).getEntry();
```
azvanderpas, line 59: Optional suggestion:

Since PIDController implements Sendable, you can actually just put
`configTab.add("Turret PID", turretPIDController);` after you instantiate turretPIDController and that will add a widget entry that updates the PID values from shuffleboard auto-magically.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+            - Constants.TurretConstants.LIMELIGHT_HEIGHT_INCHES;
+    double distance = height / Math.tan(tagTheta);
+    return distance;
+  }
+
+  @Override
+  public void periodic() {
+    turretPIDController.setP(kP.getDouble(turretPosition));
+    turretPIDController.setI(kI.getDouble(turretPosition));
+    turretPIDController.setD(kD.getDouble(turretPosition));
```
azvanderpas, line 220: If you use the PIDController widget option, don't forget to remove these.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+  public double getTurretPos() {
+    return turretPosition; // returns the absolute encoder position in radians
+  }
+
+  public double getTurretVel() {
+    return turretVelocity;
+  }
+
+  public Command testTurretCommand(double degrees) {
+    return runEnd(() -> setTurretPos(Math.toRadians(degrees)), this::stopTurret);
```
azvanderpas, line 170: Add .withName()

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+            turretPosition,
+            MathUtil.clamp(
+                setpoint,
+                Math.toRadians(-Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES),
+                Math.toRadians(Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES))));
+  }
+
+  private void holdPosition() {
+    setTurretPos(turretPosition);
+  }
```
azvanderpas, line 142: Need to set this up as default Command in RobotContainer

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+              -Math.toRadians(Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES),
+              Math.toRadians(Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES),
+              false,
+              0);
+    }
+    turretMech = new Mechanism2d(4, 4);
+    root = turretMech.getRoot("Root", 2, 2);
+    turretAngleSim =
+        root.append(new MechanismLigament2d("Turret", 2, 0, 5, new Color8Bit(Color.kRed)));
+    turretTab.add("Turret", turretMech);
```
azvanderpas, line 92: Should this be in the Robot.isSimulation if-block above? If we want it as a visual in Shuffleboard for the real robot, you'll need to update it in periodic(), not simulationPeriodic()

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -0,0 +1,246 @@
...
+
+    turretMotor.restoreFactoryDefaults();
+    turretMotor.setSmartCurrentLimit(40);
+
+    this.robotStateManager = robotStateManager;
+
+    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kReverse, -2);
+    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kForward, 2);
```
azvanderpas, line 100: The numbers here should be related to our max angle and conversion factor. 

Also verify that using negative numbers for kReverse is correct.

azvanderpas (changes requested):
Still need Commands mapped to buttons or Triggers in RobotContainer as well.

.gitignore
```diff
@@ -1,6 +1,8 @@
 # This gitignore has been specially created by the WPILib team.
 # If you remove items from this file, intellisense might break.
 
+*.txt
+
```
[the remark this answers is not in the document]
azvanderpas: Revert this since its duplicated

src/main/deploy/example.txt
azvanderpas, line 1: Need to keep example.txt, actually. Otherwise the deploy folder will be missing from git

src/main/java/frc/robot/subsystems/TurretSubsystem.java
Malachi-lewis: I don't think so. Why would we need to simulate the turret on the actual robot that has a real turret?

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -96,8 +96,8 @@ public TurretSubsystem(RobotStateManager robotStateManager) {
 
     this.robotStateManager = robotStateManager;
 
-    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kReverse, -2);
-    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kForward, 2);
+    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kReverse, (float)(-Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES / 360 * Constants.TurretConstants.CONVERSION_FACTOR));
+    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kForward, (float)(Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES / 360 * Constants.TurretConstants.CONVERSION_FACTOR));
```
azvanderpas, line 100:
```suggestion
    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kReverse, (float)(-Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES / (360 * Constants.TurretConstants.CONVERSION_FACTOR)));
    turretMotor.setSoftLimit(CANSparkMax.SoftLimitDirection.kForward, (float)(Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES / (360 * Constants.TurretConstants.CONVERSION_FACTOR)));
```

src/main/java/frc/robot/subsystems/TurretSubsystem.java
azvanderpas: Okay, the move to if-block above so it doesn't clutter Shuffleboard when we aren't simulating

azvanderpas (approved):

PR #60, opened 2024-01-19
ignore not having log file

ModBoyex:
## Justification
Git gives you an error when no log file is provided, and the code itself will prompt you for one if it can't find one. This is annoying, and the bandage for this problem was to have the boolean that you would manually change if you wanted replay mode. This fix makes it so that it catches the error for not having a log file, and just simulates normally if you don't have one.

## Testing Done
I have tested this, and it works

### DIFF
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -58,7 +58,7 @@ public class Robot extends LoggedRobot {
                   LogFileUtil.addPathSuffix(logPath, "_sim"))); // Save outputs to a new log
         }
       } catch (Exception StringIndexOutOfBoundsException) {
-        System.out.print("No log file found, simulating as normal.");
+        System.out.println("No log file found, simulating as normal.");
       }
     }
 
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -46,16 +46,20 @@ public class Robot extends LoggedRobot {
       Logger.addDataReceiver(new WPILOGWriter()); // Log to a USB stick ("/U/logs")
       Logger.addDataReceiver(new NT4Publisher()); // Publish data to NetworkTables
       new PowerDistribution(1, ModuleType.kRev); // Enables power distribution logging
-    } else if (IS_REPLAY) {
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/Robot.java
```diff
@@ -46,16 +43,20 @@ public void robotInit() {
...
-          LogFileUtil
-              .findReplayLog(); // Pull the replay log from AdvantageScope (or prompt the user)
-      Logger.setReplaySource(new WPILOGReader(logPath)); // Read replay log
-      Logger.addDataReceiver(
-          new WPILOGWriter(
-              LogFileUtil.addPathSuffix(logPath, "_sim"))); // Save outputs to a new log
+    } else {
+      try {
+        String logPath = LogFileUtil.findReplayLog();
+        if (logPath != null) {
```
azvanderpas, line 49: Checking if logPath is a valid filepath here rather than just non-null would probably avoid needing the try-catch block
java.io.File class has an exists() method that should work.

src/main/java/frc/robot/Robot.java
```diff
@@ -46,16 +43,20 @@ public void robotInit() {
...
+          setUseTiming(false); // Run as fast as possible
+
+          Logger.setReplaySource(new WPILOGReader(logPath)); // Read replay log
+          Logger.addDataReceiver(
+              new WPILOGWriter(
+                  LogFileUtil.addPathSuffix(logPath, "_sim"))); // Save outputs to a new log
+        }
+      } catch (Exception StringIndexOutOfBoundsException) {
```
azvanderpas, line 57:
```suggestion
      } catch (StringIndexOutOfBoundsException e) {
```
Should be a specific exception class you want to catch.

src/main/java/frc/robot/Robot.java
ModBoyex: I had this thought too, but the file path may be different depending on what computer is being used, and as far as I can tell, there is no way to find the file path the code is searching for.

ModBoyex: I do think that way would be better if we can figure out the file path consistantly

azvanderpas (approved):

PR #71, opened 2024-01-20
Add Intake and Chooser

Asher123456789:
Creating this as a draft PR because I still need the turret code to be pushed because I need it for photogate and making sure the angle of the turret is below 45 degrees. I also still need the correct CAN IDs for the two motors; assigned them to the temporary values of 6 & 7 for now.

### DIFF
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -76,9 +76,12 @@ public class RobotContainer {
    * joysticks}.
    */
   private void configureBindings() {
-
-    Trigger intakeButton = m_driverController.leftTrigger(0.3);
-    intakeButton.whileTrue(new IntakeCommand(intakeSubsystem));
+    Trigger turretButton = m_driverController.leftTrigger(0.3);
+    turretButton.whileTrue(new IntakeCommand(intakeSubsystem, "TURRET"));
+    Trigger trapButton = m_driverController.rightTrigger(0.3);
+    trapButton.whileTrue(new IntakeCommand(intakeSubsystem, "TRAP"));
+    Trigger reverseButton  = m_driverController.x();
+    reverseButton.whileTrue(new IntakeCommand(intakeSubsystem, "REVERSE"));
 
     // Swerve config
     drivetrain.setDefaultCommand( // Drivetrain will execute this command periodically
--- a/src/main/java/frc/robot/commands/IntakeCommand.java
+++ b/src/main/java/frc/robot/commands/IntakeCommand.java
@@ -12,14 +12,16 @@ import frc.robot.subsystems.IntakeSubsystem;
 public class IntakeCommand extends Command {
   @SuppressWarnings({"PMD.UnusedPrivateField", "PMD.SingularField"})
   private final IntakeSubsystem intakeSubsystem;
+  private String pos;
 
   /**
    * Creates a new ExampleCommand.
    *
    * @param subsystem The subsystem used by this command.
    */
-  public IntakeCommand(final IntakeSubsystem intakeSubsystem) {
+  public IntakeCommand(final IntakeSubsystem intakeSubsystem, String pos) {
     this.intakeSubsystem = intakeSubsystem;
+    this.pos = pos;
     // Use addRequirements() here to declare subsystem dependencies.
     addRequirements(intakeSubsystem);
   }
@@ -31,7 +33,15 @@ public class IntakeCommand extends Command {
   // Called every time the scheduler runs while the command is scheduled.
   @Override
   public void execute() {
-    intakeSubsystem.setIntakeSpeed(Constants.IntakeConstants.INTAKE_PERCENTAGE);
+    if (pos.equals("TURRET")) {
+      intakeSubsystem.setIntakeSpeed(Constants.IntakeConstants.INTAKE_PERCENTAGE);
+    }
+    else if (pos.equals("TRAP")) {
+      intakeSubsystem.reverseChooser(Constants.IntakeConstants.INTAKE_PERCENTAGE);
+    }
+    else {
+      intakeSubsystem.reverseIntake(Constants.IntakeConstants.INTAKE_PERCENTAGE);
+    }
   }
 
   // Called once the command ends or is interrupted.
--- a/src/main/java/frc/robot/subsystems/IntakeSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/IntakeSubsystem.java
@@ -13,11 +13,15 @@ import frc.robot.Constants;
 public class IntakeSubsystem extends SubsystemBase {
 
   private CANSparkMax intakeMotor;
+  private CANSparkMax chooserMotor;
 
   public IntakeSubsystem() {
     intakeMotor = new CANSparkMax(Constants.IntakeConstants.INTAKE_MOTOR_ID, MotorType.kBrushless);
+    chooserMotor = new CANSparkMax(Constants.IntakeConstants.INTAKE_CHOOSER_ID, MotorType.kBrushless);
     intakeMotor.restoreFactoryDefaults();
+    chooserMotor.restoreFactoryDefaults();
     intakeMotor.setSmartCurrentLimit(40);
+    chooserMotor.setSmartCurrentLimit(20);
   }
 
   public Command intakeCommand() {
@@ -29,8 +33,22 @@ public class IntakeSubsystem extends SubsystemBase {
         });
   }
 
+
+  //TODO: Add check to make sure turret is below 45 degrees before running & add photogate when implemented.
   public void setIntakeSpeed(double speed) {
     intakeMotor.set(speed);
+  }
+
+  public void reverseIntake(double speed) {
+    intakeMotor.set(-speed);
+  }
+
+  public void setChooserSpeed(double speed) {
+    chooserMotor.set(speed);
+  }
+
+  public void reverseChooser(double speed) {
+    chooserMotor.set(-speed);
   }
 
   @Override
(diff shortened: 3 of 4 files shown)


src/main/java/frc/robot/commands/IntakeCommand.java
```diff
@@ -13,13 +13,16 @@ public class IntakeCommand extends Command {
...
 
+  private String pos;
+
   /**
    * Creates a new ExampleCommand.
    *
    * @param subsystem The subsystem used by this command.
    */
-  public IntakeCommand(final IntakeSubsystem intakeSubsystem) {
+  public IntakeCommand(final IntakeSubsystem intakeSubsystem, String pos) {
```
caseyjbrotherton, line 23: Let's change this to a Command Factory in the intakesubsystem.

A command factory returns a command that can be used to assign to a trigger.  This avoids the boilerplate code.

src/main/java/frc/robot/commands/IntakeCommand.java
```diff
@@ -31,7 +34,13 @@ public void initialize() {}
   // Called every time the scheduler runs while the command is scheduled.
   @Override
   public void execute() {
-    intakeSubsystem.setIntakeSpeed(Constants.IntakeConstants.INTAKE_PERCENTAGE);
+    if (pos.equals("TURRET")) {
+      intakeSubsystem.setIntakeSpeed(Constants.IntakeConstants.INTAKE_PERCENTAGE);
+    } else if (pos.equals("TRAP")) {
+      intakeSubsystem.reverseChooser(Constants.IntakeConstants.INTAKE_PERCENTAGE);
```
caseyjbrotherton, line 40: The intakeSpeed still needs to be run when sending to the trap

Let's break this into two commands, where one runs the chooser and one runs the intake.  
We will have the chooser wheel get where it is going from the upcoming robot state management system

src/main/java/frc/robot/commands/IntakeCommand.java
lohengrin332: I'd also recommend using an `enum` to specify the position of the intake, as opposed to a string.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -76,9 +76,12 @@ public RobotContainer() {
...
   private void configureBindings() {
-
-    Trigger intakeButton = m_driverController.leftTrigger(0.3);
-    intakeButton.whileTrue(new IntakeCommand(intakeSubsystem));
+    Trigger turretButton = m_driverController.leftTrigger(0.3); 
+    turretButton.whileTrue(new StartEndCommand(intakeSubsystem::intakeCommand, intakeSubsystem::stopIntake, intakeSubsystem));
+    Trigger trapButton = m_driverController.rightTrigger(0.3);
+    trapButton.whileTrue(new StartEndCommand(intakeSubsystem::intakeCommand, intakeSubsystem::stopIntake, intakeSubsystem));
+    Trigger reverseButton = m_driverController.x();
+    reverseButton.whileTrue(new StartEndCommand(intakeSubsystem::reverseIntake, intakeSubsystem::stopIntake, intakeSubsystem));
```
KianChilds, line 84: Should use OI

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -11,28 +11,54 @@
...
+    intakeMotor.set(Constants.IntakeConstants.INTAKE_PERCENTAGE);
   }
 
-  public void setIntakeSpeed(double speed) {
-    intakeMotor.set(speed);
+  public void reverseIntake() {
+    intakeMotor.set(-Constants.IntakeConstants.INTAKE_PERCENTAGE);
   }
 
+  public void setChooseSpeed() {
```
KianChilds, line 41: A set method should take in a value

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -11,28 +11,54 @@
...
+  public void reverseChooser() {
+    chooserMotor.set(-Constants.IntakeConstants.INTAKE_PERCENTAGE);
+  }
+
+  public void stopIntake() {
+    intakeMotor.stopMotor();
+    chooserMotor.stopMotor();
+  }
+
+  public void chooserState() {
```
KianChilds, line 54: A method name should be a verb. This does not describe whatever action this is meant to perform

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -11,28 +11,54 @@
...
+  public void reverseIntake() {
+    intakeMotor.set(-Constants.IntakeConstants.INTAKE_PERCENTAGE);
   }
 
+  public void setChooseSpeed() {
+    chooserMotor.set(Constants.IntakeConstants.INTAKE_PERCENTAGE);
+  }
+
+  public void reverseChooser() {
+    chooserMotor.set(-Constants.IntakeConstants.INTAKE_PERCENTAGE);
```
KianChilds, line 46: Can we assume that the chooser and intake will run at the same percentages?

Shazamboo: we shall see.

caseyjbrotherton: Fair comment @KianChilds 

@Shazamboo Let's make two constants and set them the same.  It is likely they will simply run at the same speed

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -9,28 +9,60 @@
...
-        });
+  public void stopMotors() {
+    chooserMotor.set(0);
+    intakeMotor.set(0);
   }
 
-  public void setIntakeSpeed(double speed) {
-    intakeMotor.set(speed);
+  public Command intakeCommand(RobotStateManager robotStateManager) {
+    return robotStateManager.getPlacementMode().getAsInt() == 0
```
caseyjbrotherton, line 63: Let's not use 0 here.  

Let's use the enum from stateManagement.PlacementMode

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -76,44 +89,51 @@ public RobotContainer() {
...
-    drivetrain.registerTelemetry(logger::telemeterize);
+  }
+
+  // Register commands for auton
+  public void registerCommands() {
+    HashMap<String, Command> autonCommands = new HashMap<String, Command>();
+
+    autonCommands.put("Shoot", autonTest());
+
+    NamedCommands.registerCommands(autonCommands);
```
caseyjbrotherton, line 136: We need an Intake named command to be added as well.

caseyjbrotherton (changes requested):

src/main/java/frc/robot/stateManagement/RobotStateManager.java
```diff
@@ -52,6 +52,14 @@ public NoteState getNoteState() {
   }
 
   // Placement Mode
+  public void switchPlacementMode() {
+    if (this.placementMode.equals(PlacementMode.SPEAKER)) {
```
lohengrin332, line 56: A few things:
1. I think the `this.` prefix is unnecessary here, and you're not using it when you use `placementMode` below. This looks odd and could make people do a double-take (IOW it made me question whether this code was doing what it appears to be doing). I'd suggest removing the `this.` prefix, or if you prefer you could add it to the other uses of `placementMode` below.
2. Given that `placementMode` is an enumeration, you can compare it using `==`; you don't need `.equals()`
3. Given the above, this might be done more concisely with a ternary operator. E.g.:

`placementMode = placementMode == PlacementMode.SPEAKER ? PlacementMode.AMP : PlacementMode.SPEAKER;`

Or if you prefer to split it across multiple lines, I like this formatting for ternaries:

```
placementMode = placementMode == PlacementMode.SPEAKER
  ? PlacementMode.AMP
  : PlacementMode.SPEAKER;
```

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -10,40 +10,75 @@
...
 
-  public Command intakeCommand() {
-    // Inline construction of command goes here.
-    // Subsystem::RunOnce implicitly requires `this` subsystem.
-    return runOnce(
-        () -> {
-          /* one-time action goes here */
-        });
+  public Command reverseIntakeCommand() {
+    return new StartEndCommand(this::reverseIntake, this::stopMotors, this);
```
azvanderpas, line 63: Does the chooser need to reverse as well ? i.e. run the opposite of the current placement

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
```diff
@@ -10,40 +10,75 @@
...
-    return buildCommand(Constants.IntakeConstants.INTAKE_PERCENTAGE);
+  public Command getSpeakerIntakeCommand() {
+    return buildIntakeCommand(true);
   }
 
-  public Command getOuttakeCommand() {
-    return buildCommand(-Constants.IntakeConstants.INTAKE_PERCENTAGE);
+  public Command getAmpIntakeCommand() {
+    return buildIntakeCommand(false);
   }
```
azvanderpas, line 77: Add .withName(...) to the end of the returned Command so we can track the names in logs

src/main/java/frc/robot/subsystems/IntakeSubsystem.java
ModBoyex: Coast mode should be fine.

caseyjbrotherton (approved):

PR #73, opened 2024-01-20
Added states for endgame, note state, and placement mode

Malachi-lewis:
The rest of the robot will likely need states, so we should have them in RobotStateManager as quickly as possible.


src/main/java/frc/robot/NoteState.java
```diff
@@ -0,0 +1,28 @@
+package frc.robot;
+
+import java.util.Arrays;
+
+public enum NoteState {
+  NO_TE(0),
+  LOADING(10),
+  CHAMBER(20),
+  READY_TO_FIRE(30),
```
lohengrin332, line 9: Technically, `READY_TO_FIRE` is about more than just the note; it's about the note plus the angle of the turret. I would add this as an additional queryable state or boolean.

src/main/java/frc/robot/stateManagement/NoteState.java
```diff
@@ -0,0 +1,28 @@
+package frc.robot.stateManagement;
+
+import java.util.Arrays;
+
+public enum NoteState {
+  NO_TE(0),
```
caseyjbrotherton, line 6: What does this state mean?  I assume no note, let's make it more clear.

src/main/java/frc/robot/stateManagement/RobotStateManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public RobotStateManager() {}
+
+  @Override
+  public void periodic() {
+    if (checkAllianceColor) {
+      Optional<Alliance> alliance = DriverStation.getAlliance();
+      if (alliance.isPresent()) {
+        checkAllianceColor = false;
+        alliancePublisher.accept(alliance.get().equals(Alliance.Red) ? 10 : 20);
```
caseyjbrotherton, line 49: We should use the enum here instead of 10/20

Malachi-lewis: done

src/main/java/frc/robot/stateManagement/RobotStateManager.java
```diff
@@ -0,0 +1,71 @@
...
+import edu.wpi.first.wpilibj.DriverStation.Alliance;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import java.util.Optional;
+
+public class RobotStateManager extends SubsystemBase {
+  // Alliance Color
+  private final IntegerTopic allianceTopic =
+      NetworkTableInstance.getDefault().getIntegerTopic("ALLIANCE");
+  private final IntegerPublisher alliancePublisher = allianceTopic.publish();
+  private boolean checkAllianceColor = true;
```
caseyjbrotherton, line 19: FMS already stores the result of isRedAlliance.  I don't think we need to have another network table for alliance.  

I would expect RobotStateManager to check whether isRedAlliance in initialization, and store that in a private boolean and then return that information using a getter.

Malachi-lewis: done

src/main/java/frc/robot/stateManagement/NoteState.java
```diff
@@ -0,0 +1,28 @@
...
+  LOADING(10),
+  CHAMBER(20),
+  READY_TO_FIRE(30),
+  AMP(40);
```
caseyjbrotherton, line 10: What does AMP mean here?  Ready for amp placement?  Might want to follow the READY_TO_FIRE pattern, and put READY_TO_ in front of amp.

caseyjbrotherton (changes requested):
Left comments.  Another thought is whether the enums should be in their own class, or if they should be included in RobotStateManagement

We should think more about that.  

Also, another thought.  Do we expect the different subsystems to get the state updates directly from networktables, or from RobotStateManagement?  I assume networktables because we have a NT publisher.  But wanted to be sure.  We will need to make sure to show the team an example so that they follow the same pattern

src/main/java/frc/robot/stateManagement/NoteState.java
caseyjbrotherton: Talking to Malachi - READY_TO_PLACE?

src/main/java/frc/robot/stateManagement/RobotStateManager.java
```diff
@@ -0,0 +1,63 @@
...
+            alliance.get().equals(Alliance.Red) ? AllianceColor.RED : AllianceColor.BLUE;
+      }
+    }
+    if (!isEndGame) {
+      double matchTimer = DriverStation.getMatchTime();
+      boolean isTeleopEnabled = DriverStation.isTeleopEnabled();
+      if (matchTimer < 20 && isTeleopEnabled) {
+        isEndGame = true;
+      }
+    }
```
azvanderpas, line 38: Consider writing this using [Triggers](https://docs.wpilib.org/en/stable/docs/software/commandbased/binding-commands-to-triggers.html#arbitrary-triggers)

caseyjbrotherton (dismissed):
Changes look good

caseyjbrotherton (dismissed):

caseyjbrotherton (approved):

src/main/java/frc/robot/stateManagement/RobotStateManager.java
caseyjbrotherton: Trigger added

caseyjbrotherton (approved):

src/main/java/frc/robot/NoteState.java
caseyjbrotherton: Resolving to merge today.  @lohengrin332 let's talk about this on Tuesday and improve forward

caseyjbrotherton (approved):

PR #74, opened 2024-01-21
Trap arm

Malachi-lewis:
## Justification
Adds the basic skeleton functions of the Arm Trap that we need.


src/main/java/frc/robot/BuildConstants.java
```diff
@@ -7,12 +7,12 @@ public final class BuildConstants {
   public static final String MAVEN_GROUP = "";
```
caseyjbrotherton, line 7: This shouldn't be checked into git

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,224 @@
...
+        () -> {
+          rollerMotor.stopMotor();
+        });
+  }
+
+  public Command zeroArm() {
+    return startEnd(
+      () -> {
+        if (!baseBreak.get()) {
+          baseMotor1.getPIDController().setReference(10, ControlType.kVelocity);
```
caseyjbrotherton, line 177: Should we have a constant for this?  ( Assume this is move slowly. )

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,224 @@
...
+          wristMotor.stopMotor();
+          baseMotor1.stopMotor();
+          baseMotor2.stopMotor();
+          scoringMotor.stopMotor();
+        });
+  }
+
+  @Override
+  public void periodic() {
+    SmartDashboard.putBoolean("Source Beam Break", sourceBreak.get());
```
caseyjbrotherton, line 218: Do we want these to be DebugEntry instead?

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,224 @@
...
+    baseMotor1.getPIDController().setIZone(basePID[3]);
+    baseMotor1.getPIDController().setFF(basePID[4]);
+
+    baseMotor2 = new CANSparkMax(TrapArmConstants.baseMotor2_ID, MotorType.kBrushless);
+    baseMotor2.restoreFactoryDefaults();
+    baseMotor2.getPIDController().setP(basePID[0]);
+    baseMotor2.getPIDController().setI(basePID[1]);
+    baseMotor2.getPIDController().setD(basePID[2]);
+    baseMotor2.getPIDController().setIZone(basePID[3]);
+    baseMotor2.getPIDController().setFF(basePID[4]);
```
caseyjbrotherton, line 106: I feel like basemotor will follow each other. But not sure

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,224 @@
...
+import edu.wpi.first.wpilibj.shuffleboard.ShuffleboardTab;
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Constants.TrapArmConstants;
+
+public class TrapArmSubsystem extends SubsystemBase {
+  // Wrist motors
+  private final CANSparkMax wristMotor;
+  private final CANSparkMax rollerMotor;
```
caseyjbrotherton, line 25: Do we want to name this differently?  There are plenty of rollers on the robot

src/main/java/frc/robot/BuildConstants.java
ModBoyex: Fixed?

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
ModBoyex: I had problems when prototyping when just using follow.

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
ModBoyex: These are the only rollers for this subsystem and will be referenced in the context of the subsystem out side of the subsystem.

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
ModBoyex: done

src/main/java/frc/robot/BuildConstants.java
ModBoyex: It seems that Git will not build without the Build Constaints file.

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
ModBoyex: That might make more sense

src/main/java/frc/robot/BuildConstants.java
caseyjbrotherton: Yeah, so you can delete it, and then run build.  It will recreate the BuildConstants.java, and that's fine.  That will get sent to the robot.  But we don't want it stored in git

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
ModBoyex: done

src/main/java/frc/robot/BuildConstants.java
ModBoyex: you might be right

.gitignore
```diff
@@ -188,4 +188,4 @@ local.properties
 .vscode/settings.json
 *.wpilog
 
-BuildConstants.java
\ No newline at end of file
+src\main\java\frc\robot\BuildConstants.java
```
azvanderpas, line 191: Slashes are wrong direction

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,315 @@
...
+    scoringMotor.getPIDController().setP(scoringPID[0]);
+    scoringMotor.getPIDController().setI(scoringPID[1]);
+    scoringMotor.getPIDController().setD(scoringPID[2]);
+    scoringMotor.getPIDController().setIZone(scoringPID[3]);
+    scoringMotor.getPIDController().setFF(scoringPID[4]);
+
+    scoringBreak = new DigitalInput(TrapArmConstants.scoringBreak_ID);
+
+    // 2D Mechanism
+    armMechanism = new Mechanism2d(5, 8);
```
caseyjbrotherton, line 145: Should this be instantiated only when in simulation?

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,315 @@
...
+          5.4,
+          Units.inchesToMeters(1),
+          0,
+          Units.inchesToMeters(18),
+          true,
+          0,
+          VecBuilder.fill(0.01));
+
+    // SmartDashboard
+    sourceLog = new DebugEntry<Boolean>(baseBreak.get(), "Bace Limit Switch", this);
```
caseyjbrotherton, line 177: Typo: Base

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,358 @@
...
+              0,
+              Units.inchesToMeters(18),
+              true,
+              0);
+
+      TrapArmTab.add("Trap Arm Mech", armMechanism);
+    } else {
+      m_baseElevatorSim = null;
+      m_scoringElevatorSim = null;
+    }
```
azvanderpas, line 201: I'd prefer removing this else clause and then remove the 'final' keyword from the instantiations so it doesn't complain when they stay unassigned.

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,358 @@
...
+    baseMotor1.getPIDController().setP(basePID[0]);
+    baseMotor1.getPIDController().setI(basePID[1]);
+    baseMotor1.getPIDController().setD(basePID[2]);
+    baseMotor1.getPIDController().setFF(basePID[4]);
+
+    scoringPID = scoringPIDEntry.getDoubleArray(scoringPID);
+    scoringMotor.getPIDController().setP(scoringPID[0]);
+    scoringMotor.getPIDController().setI(scoringPID[1]);
+    scoringMotor.getPIDController().setD(scoringPID[2]);
+    scoringMotor.getPIDController().setFF(scoringPID[4]);
```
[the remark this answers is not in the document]
azvanderpas: My latest addition to the CANSparkMaxSim is to make the PIDController something you can add to Shuffleboard directly. Lets try that out.

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,315 @@
...
+          } else {
+            scoringMotor.stopMotor();
+          }
+        },
+        () -> {
+          baseMotor1.stopMotor();
+          baseMotor2.stopMotor();
+          scoringMotor.stopMotor();
+        });
+  }
```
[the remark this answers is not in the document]
ModBoyex: I did not consider that fact the PID would affect kVelocity differently, so it might just be better to just use set() rather than the PIDController

src/main/java/frc/robot/subsystems/TrapArmSubsystem.java
```diff
@@ -0,0 +1,358 @@
...
+          if (Robot.isSimulation()) {
+            wristMech.setAngle(-170);
+          }
+        });
+  }
+
+  public Command scoreTrap() {
+    return startEnd(
+        () -> {
+          setTrapArm(TrapArmState.TRAP_SCORE);
```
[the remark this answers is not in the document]
azvanderpas:
```suggestion
          setTrapArm(TrapArmState.TRAP_SCORE).initialize();
```
You could do this. And then use `setTrapArm(TrapArm.State.TRAP_SCORE).end(false);` in the end block.

src/main/java/frc/robot/BuildConstants.java
ModBoyex: fixed

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,340 @@
...
+    return startEnd(
+        () -> {
+          // Command for zering elevator if elevator happens to be not at zero
+          // Runs elevator motors until there limit switches are pressed
+          if (!baseLimit.get()) {
+            baseMotor1.set(.1);
+            baseMotor2.set(.1);
+          } else {
+            baseMotor1.stopMotor();
+            baseMotor2.stopMotor();
```
KianChilds, line 273: The motors should also be zeroed. 
Because the if the limit wired to the DIO pins as suggested by the use of DigitalInput then the SparkMax won't automatically zero

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,340 @@
...
+    private Double scoringPose;
+
+    TrapElvState(Double wrist, Double base, Double scoring) {
+      this.wristPose = wrist;
+      this.basePose = base;
+      this.scoringPose = scoring;
+    }
+
+    private double feetToRotations(double f) {
+      return Units.inchesToMeters(f) / (2 * Math.PI * Units.inchesToMeters(1)) * 70;
```
KianChilds, line 96: The constants in the equation should be moved to Constants

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    private double angleToRotations(double a) {
```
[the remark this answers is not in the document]
ModBoyex: Degrees to rotations of the motor

src/main/java/com/revrobotics/CANSparkMaxSim.java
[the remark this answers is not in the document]
ModBoyex: Yes, we are using sim alongside building out the whole system

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,374 @@
...
+    scoringLog.log(sourceBreak.get());
+  }
+
+  @Override
+  public void simulationPeriodic() {
+    for (double i = 0; i < Robot.defaultPeriodSecs; i += CANSparkMaxSim.kPeriod) {
+      m_baseElevatorSim.setInput(baseMotor1.get() * RobotController.getBatteryVoltage());
+      m_baseElevatorSim.update(CANSparkMaxSim.kPeriod);
+      baseMotor1.update(
+          m_baseElevatorSim.getVelocityMetersPerSecond() * 70 / Units.inchesToMeters(1));
```
azvanderpas, line 349:
```suggestion
          m_baseElevatorSim.getVelocityMetersPerSecond() * TrapElvConstants.elevatorGearRatio / Units.inchesToMeters(1));
```

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,374 @@
...
+      m_scoringElevatorSim.setInput(scoringMotor.get() * RobotController.getBatteryVoltage());
+      m_scoringElevatorSim.update(CANSparkMaxSim.kPeriod);
+      scoringMotor.update(
+          m_scoringElevatorSim.getVelocityMetersPerSecond()
+              * TrapElvConstants.elevatorGearRatio
+              / Units.inchesToMeters(1));
+
+      m_wristMotorSim.setInput(wristMotor.get() * RobotController.getBatteryVoltage());
+      m_wristMotorSim.update(CANSparkMaxSim.kPeriod);
+      wristMotor.update(m_wristMotorSim.getVelocityRadPerSec());
```
azvanderpas, line 360: Gear ratio needs to be factored in here I think. 

Also, I think the plan is to use the CANCoders for the wrist, so we have to use the PIDController on the roboRIO and can't integrate with the SparkMax controller. Okay to address this in a future PR though (just make a new work item so we don't forget).

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,374 @@
...
+          baseMech.append(
+              new MechanismLigament2d("Scoring Arm", 0, 0, 10, new Color8Bit(Color.kAqua)));
+      wristMech =
+          scoringMech.append(
+              new MechanismLigament2d(
+                  "Wrist Mech", TrapElvConstants.wristLength, -170, 5, new Color8Bit(Color.kRed)));
+
+      m_baseElevatorSim =
+          new ElevatorSim(
+              DCMotor.getNEO(1),
```
azvanderpas, line 185: Should this be 2 NEOs?

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,374 @@
...
+              TrapElvConstants.elevatorMinHight,
+              TrapElvConstants.elevatorMaxHight,
+              true,
+              0);
+
+      m_wristMotorSim =
+          new SingleJointedArmSim(
+              DCMotor.getNEO(1),
+              TrapElvConstants.elevatorGearRatio,
+              1,
```
azvanderpas, line 209: Not certain about the physics, but I think the moment of inertia is:
(1/3) * (wristLength^2) * elevatorCariageMass

Mike said most of the weight is the wrist arm, so using the same elevator mass seems reasonable. 

1 kg\*m^2 is actually much higher than this formula expects (~0.18 kg*m^2)

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,374 @@
...
+              Units.inchesToMeters(1),
+              TrapElvConstants.elevatorMinHight,
+              TrapElvConstants.elevatorMaxHight,
+              true,
+              0);
+
+      m_wristMotorSim =
+          new SingleJointedArmSim(
+              DCMotor.getNEO(1),
+              TrapElvConstants.elevatorGearRatio,
```
azvanderpas, line 208: I'm betting this won't be the same ratio as the elevator

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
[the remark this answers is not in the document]
azvanderpas: I think he means change the name to something like degreesToRotation(). And then maybe notice that that conversion method already exists as Units.degreesToRotations()

azvanderpas (changes requested):

azvanderpas:
Need to merge from main to fix CI build

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
ModBoyex: well, it's the same gear ratio as 1 NEO right?

azvanderpas (approved):

PR #75, opened 2024-01-21
Add metadata to gitignore

zbrotherton:
## Justification

branch.txt, commit.txt, hostname.txt are all git-tracked, when they should really just be on the robot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -188,4 +188,7 @@ local.properties
 .vscode/settings.json
 *.wpilog
 
+branch.txt
+commit.txt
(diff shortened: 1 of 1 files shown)


caseyjbrotherton:
https://github.com/frc6377/crescendo_2024/commit/02469cac39ce6eb42f8d57f6d79e07d943d592fe

Already done and merged to main.

azvanderpas:
This won't have an effect, because any branches that are producing these files aren't updated to main anyway...

PR #81, opened 2024-01-25
Shooter skeleton

aetherstorm-dev:
## Justification

Adds skeleton functionality of the shooter.


src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.revrobotics.CANSparkLowLevel.MotorType;
+import com.revrobotics.CANSparkMax;
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Constants;
+
+public class ShooterSubsystem extends SubsystemBase {
+
+  private CANSparkMax shooterMotor;
```
lohengrin332, line 11: Please make this variable `final`; it is assigned once in the constructor and then never changed again, so it is a good candidate for being made final.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,53 @@
...
+  boolean readyToFire = true; // Take in readiness state from ShooterTrigger?
+
+  public ShooterSubsystem() {
+    shooterMotor =
+        new CANSparkMax(Constants.ShooterConstants.SHOOTER_MOTOR_ID, MotorType.kBrushless);
+    shooterMotor.restoreFactoryDefaults();
+    shooterMotor.setSmartCurrentLimit(40);
+  }
+
+  public Command shooterFire(double distance) {
```
lohengrin332, line 21: Same for function parameters; please make them all `final`.

aetherstorm-dev:
TODO:
Configurable minimum "active" shooter speed, to minimize current draw
Shooter speed equation not dependent on turret angle

src/main/java/frc/robot/Constants.java
```diff
@@ -18,6 +18,11 @@ public static class IntakeConstants {
     public static final double INTAKE_PERCENTAGE = -0.75;
   }
 
+  public static class ShooterConstants {
+    public static final int SHOOTER_MOTOR_ID = 55;
+    public static final double SHOOTER_MINIMUM_SPEED = 0.1;
```
caseyjbrotherton, line 23: I think we will likely want this as an RPM. 

I am not sure.  If we decide to keep this as a motor output, we should change the variable name

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,51 @@
...
+import com.revrobotics.CANSparkLowLevel.MotorType;
+import com.revrobotics.CANSparkMax;
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import frc.robot.Constants;
+
+public class ShooterSubsystem extends SubsystemBase {
+
+  private final CANSparkMax shooterMotor;
+  boolean readyToFire = true; // Take in readiness state from ShooterTrigger?
```
caseyjbrotherton, line 12: There are several parts to being "readyToFire"  

One is having a note in the trigger, another is having the turret in position, another is the shooter rollers are at the correct speed.

Let's remove this, and instead provide a trigger that indicates whether the shooter rollers are the correct speed.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,51 @@
...
+        });
+  }
+
+  public void setShooterSpeed(double speed) {
+    shooterMotor.set(speed);
+  }
+
+  // Parameter: TargetType target ???
+  public double calculateShooterSpeed(double distance) {
+    double speed = distance; // Placeholder
```
caseyjbrotherton, line 48: Let's build out an interpolation function.

We can make up speeds for against the speaker, the first field line, end of our side and mid field.  
Find the distances from the rules.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public Command shooterFire(final double distance) {
+    return startEnd(
+        () -> {
+          if (readyToFire) {
+            readyShooter(distance);
+          }
+        },
+        () -> {
+          setShooterSpeed(Constants.ShooterConstants.SHOOTER_MINIMUM_SPEED);
```
caseyjbrotherton, line 29: Let's do this as a default command.  

https://first.wpi.edu/wpilib/allwpilib/docs/release/java/edu/wpi/first/wpilibj2/command/CommandScheduler.html#setDefaultCommand(edu.wpi.first.wpilibj2.command.Subsystem,edu.wpi.first.wpilibj2.command.Command)


So, this would be two steps:
One, a method to set the motor to the minimum speed.
Two, a call to setDEfaultCommand in RobotContainer for ShooterSubsystem.

Then we don't need to make sure to manage when we need to switch to the minimum speed.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,51 @@
...
+  public Command readyShooter(final double distance) {
+    // TargetType target =  ||| Take in target type from vision / elsewhere
+    return runOnce(
+        () -> {
+          setShooterSpeed(calculateShooterSpeed(distance)); // Take in distance from vision
+        });
+  }
+
+  public void setShooterSpeed(double speed) {
+    shooterMotor.set(speed);
```
caseyjbrotherton, line 43: We need more than setting the speed.  We need a PID to make sure that the shooter motor makes it to the right speed.

We also need a trigger for when it gets to that speed within a certain tolerance.  ( The tolerance should be in constants. )

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved with upcoming commit

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved

src/main/java/frc/robot/Constants.java
aetherstorm-dev: Resolved (placeholder value)

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved with upcoming commit (placeholder PID and tolerance)

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,69 @@
...
+    return runOnce(
+        () -> {
+          setShooterSpeed(calculateShooterSpeed(distance)); // Take in distance from vision
+        });
+  }
+
+  public boolean isShooterReady(double distance) {
+    boolean shooterReady = false;
+    double minSpeedTolerance = calculateShooterSpeed(distance)*(1-Constants.ShooterConstants.SHOOTER_SPEED_TOLERANCE);
+    double maxSpeedTolerance = calculateShooterSpeed(distance)/(1-Constants.ShooterConstants.SHOOTER_SPEED_TOLERANCE);
```
azvanderpas, line 51:
```suggestion
    double maxSpeedTolerance = calculateShooterSpeed(distance)*(1+Constants.ShooterConstants.SHOOTER_SPEED_TOLERANCE);
```

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved in commit b99820f.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved in commit 13a1856.

src/main/java/frc/robot/Constants.java
```diff
@@ -18,6 +18,13 @@ public static class IntakeConstants {
     public static final double INTAKE_PERCENTAGE = -0.75;
   }
 
+  public static class ShooterConstants {
+    public static final int SHOOTER_MOTOR_ID = 55;
+    public static final double SHOOTER_MINIMUM_SPEED = 150; // Placeholder; in RPM
```
KianChilds, line 23: `SHOOTER_IDLE_SPEED` is a more descriptive name

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+        });
+  }
+
+  // Checks if shooter is ready.
+  public boolean isShooterReady(double distance) {
+    boolean shooterReady = false;
+    double minSpeedTolerance =
+        SpeakerRanges.getSpeed(distance) * (1 - Constants.ShooterConstants.SHOOTER_SPEED_TOLERANCE);
+    double maxSpeedTolerance =
+        SpeakerRanges.getSpeed(distance) * (1 + Constants.ShooterConstants.SHOOTER_SPEED_TOLERANCE);
```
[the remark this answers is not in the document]
aetherstorm-dev: Resolved (Constants.java was altered to match this)

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+      {195, 750},
+      {290, 1000}
+    };
+
+    public static double getSpeed(double distance) {
+      double distanceProportion;
+      // A linear search which determines which points the input distance falls between. May be
+      // converted to a binary search if there are many points
+      for (int i = 0; i <= SpeakerRanges.length; i++) {
+        if (i == SpeakerRanges.length) {
```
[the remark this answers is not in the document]
aetherstorm-dev: Resolved, dumb mistake on my end lol

src/main/java/frc/robot/Constants.java
aetherstorm-dev: Resolved in 429dcbb

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+    shooterMotor =
+        new CANSparkMax(Constants.ShooterConstants.SHOOTER_MOTOR_ID, MotorType.kBrushless);
+    shooterMotor.restoreFactoryDefaults();
+    shooterMotor.setSmartCurrentLimit(40);
+
+    // Placeholder values
+    shooterMotor.getPIDController().setP(0);
+    shooterMotor.getPIDController().setI(0);
+    shooterMotor.getPIDController().setD(0);
+    shooterMotor.getPIDController().setFF(0);
```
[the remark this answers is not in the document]
aetherstorm-dev: Resolved in 429dcbb (moved to Constants.ShooterConstants)

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+  // Speed in RPM.
+  public void setShooterSpeed(double speed) {
+    shooterMotor.getPIDController().setReference(speed, CANSparkBase.ControlType.kVelocity);
+  }
+
+  private class SpeakerRanges {
+    private static double speed;
+
+    // Placeholder; in inches and RPM respectively.
+    private static double[][] SpeakerRanges = {
```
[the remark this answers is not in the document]
aetherstorm-dev: Resolved in 429dcbb (successor variable made camelcase)

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+      {40, 550},
+      {195, 750},
+      {290, 1000}
+    };
+
+    public static double getSpeed(double distance) {
+      double distanceProportion;
+      // A linear search which determines which points the input distance falls between. May be
+      // converted to a binary search if there are many points
+      for (int i = 0; i <= SpeakerRanges.length; i++) {
```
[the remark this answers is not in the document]
aetherstorm-dev: Resolved in 429dcbb

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
KianChilds, line 1: Needs to add support for split roller set up

aetherstorm-dev: 2x SparkMax? Do the motors / rollers run the same way?

caseyjbrotherton: Run the same way, but not at the same speed.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,125 @@
...
+
+  // Fires the shooter.
+  public Command shooterFire(double distance) {
+    return startEnd(
+        () -> {
+          if (isShooterReady(distance)) {
+            setShooterSpeed(calculateShooterSpeed(distance));
+          }
+        },
+        () -> {
```
caseyjbrotherton, line 35: Let's remove code that is commented out

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Are the speeds linearly proportional? If so, what is the proportionality constant?

aetherstorm-dev: Acknowledged in 0995d8d

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved in 0995d8d

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,152 @@
...
+        calculateShooterSpeeds(distance)[0]
+            * (1 + Constants.ShooterConstants.SHOOTER_SPEED_TOLERANCE);
+
+    if (minSpeedTolerance < shooterTopMotor.getEncoder().getVelocity()
+        & shooterTopMotor.getEncoder().getVelocity() < maxSpeedTolerance) {
+      shooterReady = true;
+    }
+
+    return shooterReady;
+  }
```
azvanderpas, line 71:
```suggestion
    double target = calculateShooterSpeeds(distance)[0];
    double minSpeedTolerance = target * (1 - Constants.ShooterConstants.SHOOTER_SPEED_TOLERANCE);
    double maxSpeedTolerance = target * (1 + Constants.ShooterConstants.SHOOTER_SPEED_TOLERANCE);

    double vel = shooterTopMotor.getEncoder().getVelocity();
    return (minSpeedTolerance < vel && vel < maxSpeedTolerance);
  }
```

Avoid doing complicated methods (e.g. reading a sensor or interpolation calculations) multiple times when the values are expected to be the same.

& is a bitwise AND operator, which probably winds up being the same result in this case, but most often you want the && logical AND operator.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,152 @@
...
+    shooterBottomMotor.getPIDController().setI(Constants.ShooterConstants.SHOOTER_I);
+    shooterBottomMotor.getPIDController().setD(Constants.ShooterConstants.SHOOTER_D);
+    shooterBottomMotor.getPIDController().setFF(Constants.ShooterConstants.SHOOTER_FF);
+  }
+
+  // Fires the shooter.
+  public Command shooterFire(double distance) {
+    return startEnd(
+        () -> {
+          if (isShooterReady(distance)) {
```
azvanderpas, line 40: Logic seems inverted.

azvanderpas (commented):
Consider using Pair<Double,Double> class or another data structure instead of a double array to pass around you bottom and top motor speeds. Then there's less risk of having an incorrectly allocated array input crashing the robot.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,152 @@
...
+        () -> {});
+  }
+
+  // Idle shooter command; for default command purposes
+  public Command shooterIdle() {
+    return run(
+        () -> {
+          setShooterSpeeds(Constants.ShooterConstants.SHOOTER_IDLE_SPEEDS);
+        });
+  }
```
azvanderpas, line 53: Append .withName(...) to the returned Commands so we can track them in logs

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Got it, committed and resolved

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,152 @@
...
+    if (minSpeedTolerance < shooterTopMotor.getEncoder().getVelocity()
+        & shooterTopMotor.getEncoder().getVelocity() < maxSpeedTolerance) {
+      shooterReady = true;
+    }
+
+    return shooterReady;
+  }
+
+  // Speed in RPM. Top is index 0, bottom is index 1.
+  public void setShooterSpeeds(double[] speeds) {
```
lohengrin332, line 66: I dislike representing non-sequential data in an array like this, because it is unnecessarily confusing. Instead, I think it would be better to either accept separate values for the upper vs lower motor, or accept a custom class which contains both. Given that this is a simple method which accepts only two parameters, it probably makes more sense to pass them in separately.

lohengrin332: However given that the motor speeds need to be tied together, it might make sense to use a custom class or an immutable.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,144 @@
...
+                  + speakerConfigList[i].getSpeedBottom();
+          break;
+        }
+      }
+    }
+    return speeds;
+  }
+
+  private static class SpeakerConfig {
+    private double distance;
```
lohengrin332, line 115: It would be useful to have a unit attached to these fields, e.g. if the distance is in feet `distanceInFeet` and if the speed is in rotations per minute `speedTopRPM`.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: The issue is that it is impossible to _return_ multiple values, which is necessary in calculateShooterSpeeds(). However, I will use a custom class as recommended by Adam.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved in 6372054

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Acknowledged with 6372054; framework exists now, cannot be fully resolved yet

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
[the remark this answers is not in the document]
aetherstorm-dev: Resolved in 6372054

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,152 @@
...
+      speeds[1] = speakerConfigList[0].getSpeedBottom();
+    } else {
+      // A linear search which determines which points the input distance falls between. May be
+      // converted to a binary search if there are many points
+      for (int i = 0; i < speakerConfigList.length; i++) {
+        // If distance above maximum, set speed to maximum.
+        if (i == speakerConfigList.length - 1) {
+          speeds[0] = speakerConfigList[i].getSpeedTop();
+          speeds[1] = speakerConfigList[i].getSpeedBottom();
+          break;
```
lohengrin332, line 98: IMO these `break` statements are unnecessary. Instead consider just `return`ing at these points in the flow.

If you do that, you could then have the loop go up to length-1 the pull this `if` block out of the loop, which I think would be more readable.

lohengrin332 (changes requested):
I have a few suggestions as comments in the code.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Acknowledged in 5def812 and a1e8952, using Pair<Double, Double>

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Doing so does not simplify the logic or, really, make it more readable; you're just shifting the code around. It just ends up adding more, similarly arcane lines in the middle of the method.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved in 5967b58

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: "Resolved" unless this change is viewed as critical

caseyjbrotherton (changes requested):
Use ShooterSpeeds instead of Pair

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,207 @@
...
+    Double bottomSpeed = 0d;
+    double distanceProportion;
+
+    // If distance below minimum, set speed to minimum.
+    if (distance < speakerConfigList[0].getDistanceInInches()) {
+      topSpeed = speakerConfigList[0].getSpeedTopInRPM();
+      bottomSpeed = speakerConfigList[0].getSpeedBottomInRPM();
+      speeds = new Pair(topSpeed, bottomSpeed);
+      return speeds;
+    } else if (distance >= speakerConfigList[0].getDistanceInInches()
```
lohengrin332, line 122: Given that this if is the opposite of the above if, you don't need this if. Also, since the above if returns at the end, you don't need the else. That would leave this line with:

```suggestion
    }
    if (distance <= speakerConfigList[speakerConfigList.length - 1].getDistanceInInches()) {
```

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,207 @@
...
+        && distance <= speakerConfigList[speakerConfigList.length - 1].getDistanceInInches()) {
+      // A linear search which determines which points the input distance falls between. May be
+      // converted to a binary search if there are many points
+      for (int i = 0; i < speakerConfigList.length - 1; i++) {
+        // If distance above maximum, set speed to maximum.
```
lohengrin332, line 126: I believe this comment applies to the block of code after this if block, currently starting on line 152.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
lohengrin332: Finally, I'd have the below for loop stop at length -2, with a good comment as to why this is being done, since it's an unusual ting to do.

aetherstorm-dev: Resolved in 47b8ded (loop already stops at length-2).

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
aetherstorm-dev: Resolved in 47b8ded.

aetherstorm-dev:
> Use ShooterSpeeds instead of Pair

Resolved in d35835b.

src/main/java/frc/robot/subsystems/ShooterSubsystem.java
```diff
@@ -0,0 +1,210 @@
...
+    // If distance below minimum, set speed to minimum.
+    if (distance < speakerConfigList[0].getDistanceInInches()) {
+      topSpeed = speakerConfigList[0].getSpeedTopInRPM();
+      bottomSpeed = speakerConfigList[0].getSpeedBottomInRPM();
+      speeds = new SpeakerConfig(distance, topSpeed, bottomSpeed);
+      return speeds;
+    }
+    // A linear search which determines which points the input distance falls between. May be
+    // converted to a binary search if there are many points
+    for (int i = 0; i < speakerConfigList.length - 1; i++) {
```
lohengrin332, line 124: You should add a comment calling out the fact that you're looping through all but the last element here. You might even consider changing it to `i <= length - 2` to make it more obvious that you're doing an un-conventional loop.

aetherstorm-dev: Resolved in ccf2513.

caseyjbrotherton (approved):

lohengrin332 (approved):
Looks good to me, thanks for the changes.

I will merge this shortly.

PR #95, opened 2024-01-28
Feature/pod zeroing

### LINKED ISSUE #94: Make Swerve Pod Zeroing Command
Add a command to Zero the DriveTrain without needing to use Phoenix Tuner or hand copy values

KianChilds:
## Justification
Fixes #94 
To make pod zeroing a easier procedure.

## Implementaion
Creates a command to automatically set the current pod rotation as zero.

## Testing Done
None

### DIFF
--- a/src/main/java/frc/robot/config/DynamicRobotConfig.java
+++ b/src/main/java/frc/robot/config/DynamicRobotConfig.java
@@ -51,6 +51,7 @@ public class DynamicRobotConfig {
   }
 
   public static void saveConfig() {
+    System.out.println("-------- SAVE!!!! --------");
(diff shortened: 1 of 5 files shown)


src/main/java/frc/robot/config/DynamicRobotConfig.java
```diff
@@ -12,7 +12,7 @@
  * IDs, Conversions, or setpoint information in this config.
  */
 public class DynamicRobotConfig {
-  private static class ConfigVariables {
+  public static class ConfigVariables {
```
azvanderpas, line 15: Making class variables public is bad practice

PR #101, opened 2024-01-31
Trap elv use CANCoder

ModBoyex:
## Justification
The wrist of the elv will have a can coder and we would like to use this rather than the internal encoder of the motor. This is because it is more precise.


src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,420 @@
...
+              DCMotor.getNEO(1),
+              TrapElvConstants.WRIST_GEAR_RATIO,
+              (1.0 / 3.0)
+                  * TrapElvConstants.ELV_LIFT_MASS
+                  * Math.pow(TrapElvConstants.WRIST_LENGTH, 2.0),
+              TrapElvConstants.WRIST_LENGTH,
+              TrapElvConstants.WRIST_MIN_ANGLE, // min rotation
+              TrapElvConstants.WRIST_MAX_ANGLE, // max rotation
+              false,
+              0);
```
azvanderpas, line 234: Gravity for this physics model points in the -90 degree direction. 

For the mechanism, however, down is 180 degrees. 

Should be able to make those agree by adding offsets in the right places.

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -0,0 +1,420 @@
...
+  public Command scoreTrap() {
+    return startEnd(
+        () -> {
+          setTrapArm(TrapElvState.TRAP_SCORE);
+          rollerMotor.set(TrapElvConstants.ROLLER_INTAKE_SPEED);
+        },
+        () -> {
+          stowTrapElv();
+        });
+  }
```
azvanderpas, line 317: Append .withName(...) to the returned Commands so we can track them in logs

riversry (approved):

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -379,18 +407,19 @@ public void simulationPeriodic() {
...
+    m_wristMotorSim.update(Robot.defaultPeriodSecs);
+    simWristPos.set(Units.radiansToRotations(m_wristMotorSim.getAngleRads()));
+
     SmartDashboard.putNumber("base CAN Sim", baseMotor1.get());
     SmartDashboard.putNumber("scoring CAN Sim", scoringMotor.get());
 
     baseMech.setLength(m_baseElevatorSim.getPositionMeters());
     scoringMech.setLength(m_scoringElevatorSim.getPositionMeters());
-    wristMech.setAngle(Units.radiansToDegrees(m_wristMotorSim.getAngleRads()));
+    wristMech.setAngle(Units.radiansToDegrees(m_wristMotorSim.getAngleRads()) - 90);
```
azvanderpas, line 422: Add comment about why the 90 degree offset is necessary.

azvanderpas (approved):

azvanderpas (approved):

PR #107, opened 2024-02-02
Add Simpler Telemetry function

ModBoyex:
## Justification
The old function was running too much and was causing loop over runs.

## Implementaion
Now we have two functions. On (The original one) that runs only if the robot is a simulation, and the new on that logs only the robot pose while the robot is real.

### DIFF
--- a/src/main/java/frc/robot/Telemetry.java
+++ b/src/main/java/frc/robot/Telemetry.java
@@ -117,7 +117,7 @@ public class Telemetry {
 
   public void realTelemetry(SwerveDriveState state) {
     double currentTime = Utils.getCurrentTimeSeconds();
-    if (currentTime - lastTime > Constants.Telemetry_Log_Number) {
+    if (currentTime - lastTime > Constants.TELEMETRY_LOG_NUMBER) {
       Pose2d pose = state.Pose;
       fieldTypePub.set("Field2d");
       fieldPub.set(new double[] {pose.getX(), pose.getY(), pose.getRotation().getDegrees()});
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/Telemetry.java
```diff
@@ -114,4 +114,13 @@ public void telemeterize(SwerveDriveState state) {
       SmartDashboard.putData("Module " + i, m_moduleMechanisms[i]);
     }
   }
+
+  public void realTelemetry(SwerveDriveState state) {
+    double currentTime = Utils.getCurrentTimeSeconds();
+    if (currentTime - lastTime > 0.2) {
```
caseyjbrotherton, line 120: Let's make this a constant.

caseyjbrotherton (changes requested):

src/main/java/frc/robot/Telemetry.java
```diff
@@ -114,4 +114,13 @@ public void telemeterize(SwerveDriveState state) {
       SmartDashboard.putData("Module " + i, m_moduleMechanisms[i]);
     }
   }
+
+  public void realTelemetry(SwerveDriveState state) {
+    double currentTime = Utils.getCurrentTimeSeconds();
+    if (currentTime - lastTime > Constants.Telemetry_Log_Number) {
```
azvanderpas, line 120: lastTime isn't updated

caseyjbrotherton (approved):

caseyjbrotherton (approved):

PR #110, opened 2024-02-03
Trap elv intake integration

ModBoyex:
## Justification

Trap Elv needs to intake at the same time as the intake subsystem to get the piece for scoring the trap.

## Implementaion

This implementation just intakes the Trap Elv when the intake is run, but it would be better to find a way to run only if the driver wants to score in the trap, and still use the same button.


src/main/java/frc/robot/stateManagement/PlacementMode.java
```diff
@@ -4,7 +4,8 @@
 
 public enum PlacementMode {
   SPEAKER(0),
-  AMP(10);
+  AMP(10),
+  SOURCE(20);
```
lohengrin332, line 8: I can see why we would want to have all of these states as within a single enumeration, but maybe `PlacementMode` is no longer an accurate name.

ModBoyex: Is that something we want to change in this PR?

azvanderpas: NodeMode?

PR #120, opened 2024-02-04
Remove CANsparkmaxSIM logic for trap arm

caseyjbrotherton:
## Justification
RObot testing on current main was showing a very delayed drive control. 
( a noticeable second between joystick axis change and robot change )

## Implementation

Disabled limelight
Did binary search through commits to discover which commit changed the behavior
Checked the periodic, simulation and button mappings.  
Nothing resolved the issue except for changing CANSparkMaxSIM to a regular CANSparkMax

## Testing Done

On robot testing as above

@zbrotherton drove this change.
Adding it in a pull request to isolate the change and allow for re-enabling

### DIFF
--- a/src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
@@ -8,7 +8,6 @@ import com.ctre.phoenix6.hardware.CANcoder;
 import com.revrobotics.CANSparkBase.ControlType;
 import com.revrobotics.CANSparkLowLevel.MotorType;
 import com.revrobotics.CANSparkMax;
-import com.revrobotics.CANSparkMaxSim;
 import edu.wpi.first.hal.SimDouble;
 import edu.wpi.first.math.MathUtil;
 import edu.wpi.first.math.controller.PIDController;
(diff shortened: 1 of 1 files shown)


src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -149,14 +147,14 @@ public TrapElvSubsystem() {
...
-    baseMotor1 = new CANSparkMaxSim(TrapElvConstants.BASE_MOTOR1_ID, MotorType.kBrushless);
+    baseMotor1 = new CANSparkMax(TrapElvConstants.BASE_MOTOR1_ID, MotorType.kBrushless);
     baseMotor1.restoreFactoryDefaults();
     baseMotor1.getPIDController().setP(TrapElvConstants.BASE_PID[0]);
     baseMotor1.getPIDController().setI(TrapElvConstants.BASE_PID[1]);
     baseMotor1.getPIDController().setD(TrapElvConstants.BASE_PID[2]);
     baseMotor1.getPIDController().setIZone(TrapElvConstants.BASE_PID[3]);
     baseMotor1.getPIDController().setFF(TrapElvConstants.BASE_PID[4]);
-    TrapElvTab.add("Base Elv PID", baseMotor1.getPIDController());
+    // TrapElvTab.add("Base Elv PID", baseMotor1.getPIDController());
```
azvanderpas, line 157: I suspect this is what was causing the issue. 
 
I didn't make initSendable() well to take into account that getP/I/D() would be making CAN transactions every cycle.

https://github.com/frc6377/crescendo_2024/blob/main/src%2Fmain%2Fjava%2Fcom%2Frevrobotics%2FCANSparkMaxSim.java#L56-L79

caseyjbrotherton: Want to test it this week?  and make a decision of next steps from that?

caseyjbrotherton:
canceling, Adam has a pull request to fix

PR #125, opened 2024-02-07
Tof sensors

ModBoyex:
## Justification

We want to be able to simply use the TOF sensor as a TOF or Beam Break sensor

## Implementaion

Creates a class for the sensor that has getters that can work for a TOF or Beam Break

## Testing Done

N/A

### DIFF
--- a/src/main/java/frc/robot/utilities/TOFSensorSimple.java
+++ b/src/main/java/frc/robot/utilities/TOFSensorSimple.java
@@ -6,6 +6,9 @@ package frc.robot.utilities;
 
 import com.playingwithfusion.TimeOfFlight;
 import edu.wpi.first.util.sendable.SendableBuilder;
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.button.Trigger;
+import java.util.function.BooleanSupplier;
 
 /** Add your docs here. */
 public class TOFSensorSimple {
@@ -21,8 +24,12 @@ public class TOFSensorSimple {
     return Sensor.getRange();
   }
 
-  public boolean isBeamBroke() {
-    return getDistance() < threshold;
+  public BooleanSupplier isBeamBroke() {
+    return () -> (getDistance() < threshold);
+  }
+
+  public Trigger getBeamTrigger(Command action) {
+    return new Trigger(isBeamBroke());
   }
 
   public void identify() {
--- a/src/main/java/frc/robot/utilities/TOFSensorSimple.java
+++ b/src/main/java/frc/robot/utilities/TOFSensorSimple.java
@@ -15,15 +15,15 @@ public class TOFSensorSimple {
 
   public TOFSensorSimple(int ID, double threshold) {
     sensor = new TimeOfFlight(ID);
-    this.threshold = threshold;
+    this.threshold = threshold; // in mm
   }
 
-  public double getDistance() {
+  public double getMillameters() {
     return sensor.getRange();
   }
 
   public boolean isBeamBroke() {
-    return getDistance() < threshold;
+    return getMillameters() < threshold;
   }
 
   public Trigger beamBroken(Command action) {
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/utilities/TOFSensorSimple.java
```diff
@@ -0,0 +1,42 @@
...
+    return Sensor.getRange();
+  }
+
+  public BooleanSupplier isBeamBroke() {
+    return () -> (getDistance() < threshold);
+  }
+
+  public Trigger getBeamTrigger(Command action) {
+    return new Trigger(isBeamBroke());
+  }
```
azvanderpas, line 33:
```suggestion
  public boolean isBeamBroke() {
    return getDistance() < threshold;
  }

  public Trigger beamBroken() {
    return new Trigger(this::isBeamBroke);
  }
```

You have the right idea, just the syntax is a bit tricky. Keeping isBeamBroke() returning a boolean makes it most useful to other classes, and the :: syntax can easily make it into a BooleanSupplier to make the Trigger

src/main/java/frc/robot/utilities/TOFSensorSimple.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.playingwithfusion.TimeOfFlight;
+import edu.wpi.first.util.sendable.SendableBuilder;
+import edu.wpi.first.wpilibj2.command.Command;
+import edu.wpi.first.wpilibj2.command.button.Trigger;
+import java.util.function.BooleanSupplier;
+
+/** Add your docs here. */
+public class TOFSensorSimple {
+  private TimeOfFlight Sensor;
```
azvanderpas, line 15: Make Sensor lower case so it doesn't get confused for a Class name

src/main/java/frc/robot/utilities/TOFSensorSimple.java
```diff
@@ -0,0 +1,42 @@
...
+    return () -> (getDistance() < threshold);
+  }
+
+  public Trigger getBeamTrigger(Command action) {
+    return new Trigger(isBeamBroke());
+  }
+
+  public void identify() {
+    Sensor.identifySensor();
+  }
```
azvanderpas, line 37: Does this blink the sensor? If so, maybe name our method blink()

src/main/java/frc/robot/utilities/TOFSensorSimple.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public void putBuilder(SendableBuilder builder) {
+    Sensor.initSendable(builder);
+  }
```
azvanderpas, line 41: I don't think this is what you want to do. 

initSendable() is so Smartdashboard or Shuffleboard know how to add an object. But our code shouldn't call it directly. 

Instead you can do something like:
` myShuffleboardTab.add("ToF Unique Name", this);` 
in the constructor, or add a method to facilitate setting this up if we don't want every ToF on NetworkTables.

azvanderpas (changes requested):
Also add comments about units for threshold/range. Or better yet include them as part of your method naming

src/main/java/frc/robot/utilities/TOFSensorSimple.java
ModBoyex: It might just be better for the subsystems that have the sensors to do that themselves

src/main/java/frc/robot/utilities/TOFSensorSimple.java
```diff
@@ -0,0 +1,36 @@
...
+public class TOFSensorSimple {
+  private TimeOfFlight sensor;
+  private double threshold;
+
+  public TOFSensorSimple(int ID, double threshold) {
+    sensor = new TimeOfFlight(ID);
+    this.threshold = threshold; // in mm
+  }
+
+  public double getMillameters() {
```
azvanderpas, line 21:
```suggestion
  public double getMilliMeters() {
```

azvanderpas (approved):

PR #126, opened 2024-02-08
Make shoot command mapped to a trigger

Josh-Westbrook:
## Justification
 Make the bot able to shoot

### DIFF
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -146,8 +146,8 @@ public class RobotContainer {
             shooterSubsystem
                 .shooterFire()
                 .until(shooterSubsystem.shooterReady())
-                .andThen(triggerSubsystem.getShootCommand()));
-
+                .andThen(triggerSubsystem.getShootCommand())
+                .onlyWhile(OI.getTrigger(OI.Operator.shooterTrigger)));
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -141,8 +141,13 @@ private void configureBindings() {
...
     shooterSubsystem.setDefaultCommand(shooterSubsystem.shooterIdle());
-    OI.getTrigger(OI.Operator.shooterTrigger).onTrue(shooterSubsystem.shooterFire());
-
+    OI.getTrigger(OI.Operator.shooterTrigger)
+        .onTrue(
+            shooterSubsystem
+                .shooterFire()
+                .until(shooterSubsystem.shooterReady())
+                .andThen(triggerSubsystem.getShootCommand())
+                .onlyWhile(OI.getTrigger(OI.Operator.shooterTrigger)));
```
azvanderpas, line 150: We should think about this some more. I think we want the functionality to be like this:
"While trigger is held, spin up shooter to desired RPM. If shooter is ready when trigger is released, shoot the note." 
Currently this code says "shoot as soon as shooter is ready", which isn't quite what we want, I don't think.

aetherstorm-dev:
This branch has been merged into ShooterSimulation for further work

PR #128, opened 2024-02-09
Dissable elv

ModBoyex:
## Justification

Robot V1 will not have an Elv, so I temp disabled the Elv.

## Implementaion

Just used If statements. This is because it's cleaner the commenting out code, it can be easily turned back on for sim, and so that it's easy to revert to normal.

## Testing Done

I've varified that this works in sim

### DIFF
--- a/src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
@@ -336,51 +336,57 @@ public class TrapElvSubsystem extends SubsystemBase {
   }
 
   public Command zeroArm() {
-    return startEnd(
-            () -> {
-              // Command for zeroing elevator if elevator happens to be not at zero
-              // Runs elevator motors until there limit switches are pressed
-              if (!baseLimit.get()) {
-                baseMotor1.set(TrapElvConstants.ELV_ZEROING_SPEED);
-                baseMotor2.set(TrapElvConstants.ELV_ZEROING_SPEED);
-              } else {
+    if (isElv) {
+      return startEnd(
+              () -> {
+                // Command for zeroing elevator if elevator happens to be not at zero
+                // Runs elevator motors until there limit switches are pressed
+                if (!baseLimit.get()) {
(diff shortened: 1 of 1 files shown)


src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -394,42 +406,46 @@ public void periodic() {
...
+                * TrapElvConstants.ELV_GEAR_RATIO
+                / TrapElvConstants.DRUM_RADIUS);
+
+        m_scoringElevatorSim.setInput(scoringMotor.get() * RobotController.getBatteryVoltage());
+        m_scoringElevatorSim.update(CANSparkMaxSim.kPeriod);
+        scoringMotor.update(
+            m_scoringElevatorSim.getVelocityMetersPerSecond()
+                * TrapElvConstants.ELV_GEAR_RATIO
+                / TrapElvConstants.DRUM_RADIUS);
+      }
```
caseyjbrotherton, line 424: Do we need to comment out simulation?  Or can we leave simulation in?

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -332,59 +336,67 @@ public void stowTrapElv() {
...
-            () -> {
-              baseMotor1.stopMotor();
-              baseMotor2.stopMotor();
-              scoringMotor.stopMotor();
-            })
-        .withName("Zero Arm");
+              })
+          .withName("Zero Arm");
+    } else {
+      return run(() -> {});
```
caseyjbrotherton, line 367: Trying to think through the pros and cons of returning an empty command vs disabling this in the robot.

Have we tested this in simulation?  Will a blank command cause any problems?  It is certainly an edge case.

caseyjbrotherton: @ModBoyex tested this button in sim.  CLosing comment

caseyjbrotherton (approved):

PR #140, opened 2024-02-14
Better sim

ModBoyex:

### DIFF
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -101,7 +101,7 @@ public final class Constants {
     // P, I, D, Iz, FF
     public static final double[] BASE_PID = {36e-3, 5e-7, 1e-4, 0.0, 2e-6};
     public static final double[] SCORING_PID = {36e-3, 5e-7, 1e-4, 0.0, 2e-6};
-    public static final double[] WRIST_PID = {7, .1, .6, 0.0, 2e-6};
+    public static final double[] WRIST_PID = {36e-3, 5e-7, 1e-4, 0.0, 2e-6};
 
     // Simulation
     public static final int ELV_GEAR_RATIO = 70;
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -72,11 +72,11 @@ public final class Constants {
     public static final double KMAXOUTPUT = 1;
     public static final double KMINOUTPUT = -1;
     public static final int MAX_TURRET_ANGLE_DEGREES = 110;
-    public static final double GEAR_RATIO = 6.8333; // 0.25 for the revbot prototype turret
+    public static final double GEAR_RATIO = 6.8333; // (24:164)  0.25 for the revbot prototype turret
 
     // Simulation Vals
     public static final double TURRET_MOI = 0; // Moment of Inertia
-    public static final double TURRET_RADIUS = 0;
+    public static final double TURRET_RADIUS = Units.inchesToMeters(12);
   }
 
   public static class OperatorConstants {
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -118,7 +118,7 @@ public final class Constants {
     public static final double WRIST_MIN_ANGLE = Units.degreesToRadians(-359); // RADS
     public static final double WRIST_MAX_ANGLE = Units.degreesToRadians(359); // RADS
     public static final double WRIST_LENGTH = Units.inchesToMeters(11.877934);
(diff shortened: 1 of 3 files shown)


src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -425,13 +436,15 @@ public void simulationPeriodic() {
     }
 
     // Wrist Sim Stuff
-    m_wristMotorSim.setInput(wristMotor.get() * RobotController.getBatteryVoltage());
+    m_wristMotorSim.setInput(wristMotor.getAppliedOutput() * RobotController.getBatteryVoltage());
```
azvanderpas, line 439: I think getAppliedOutput() returns the voltage applied, so multiplying by the battery voltage sets the input to 12x what it should be

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -397,11 +402,17 @@ public void periodic() {
       baseLog.log(baseLimit.get());
       scoringLog.log(sourceBreak.get());
     }
-    wristMotor.set(
+
+    FF =
+        wristFeedforward.calculate(
+            Units.rotationsToRadians(wristEncoder.getPosition().getValueAsDouble()), 0);
```
azvanderpas, line 408: I think the velocity shouldn't be zero. We can get the current velocity from the encoder. 
For simulation this requires setting up a new SimDouble for the CANcoder SimDevice

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -134,9 +139,15 @@ public TrapElvSubsystem() {
             TrapElvConstants.WRIST_PID[0],
             TrapElvConstants.WRIST_PID[1],
             TrapElvConstants.WRIST_PID[2]);
+    wristPIDController.setIZone(TrapElvConstants.WRIST_PID[3]);
+    wristFeedforward =
+        new ArmFeedforward(
+            TrapElvConstants.WRIST_FF[0],
+            TrapElvConstants.WRIST_FF[1] / 2.0,
```
azvanderpas, line 146: Don't think we meant to leave the divide by 2.0 here

src/main/java/frc/robot/Constants.java
```diff
@@ -72,7 +72,12 @@ public static class TurretConstants {
     public static final double KMAXOUTPUT = 1;
     public static final double KMINOUTPUT = -1;
     public static final int MAX_TURRET_ANGLE_DEGREES = 110;
-    public static final double CONVERSION_FACTOR = 0.25; // for the revbot prototype turret
+    public static final double GEAR_RATIO =
+        6.8333; // (24:164)  0.25 for the revbot prototype turret
+
```
azvanderpas, line 77: The way CONVERSION_RATIO was used here was actually the inverse of the gear ratio. So changing it to GEAR_RATIO requires inverting all the math where it is used

azvanderpas (changes requested):
Resolve conflicts with main first, since they may be fairly significant

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
ModBoyex: It does need to be half because it is still only one motor driving rotation and not two, but I will make this change in Constaints rather than here

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -425,13 +440,15 @@ public void simulationPeriodic() {
...
+    m_wristMotorSim.setInput(wristMotor.getAppliedOutput());
     m_wristMotorSim.update(Robot.defaultPeriodSecs);
     simWristPos.set(Units.radiansToRotations(m_wristMotorSim.getAngleRads()));
+    simWristVel.set(Units.radiansToRotations(m_wristMotorSim.getVelocityRadPerSec()));
     // Offest added so that gravity is simulated in the right direction
     wristMech.setAngle(Units.radiansToDegrees(m_wristMotorSim.getAngleRads()) - 90);
+    SmartDashboard.putNumber("Current Draw Wrist (A)", m_wristMotorSim.getCurrentDrawAmps());
 
-    SmartDashboard.putNumber("Wrist Motor Sim Output", wristMotor.get());
+    SmartDashboard.putNumber("Wrist Motor Sim Output", wristMotor.getAppliedOutput());
```
azvanderpas, line 451: This would be good to log in periodic, not just in simulation.

ModBoyex: Fixed in the Testing branch

PR #142, opened 2024-02-14
PhotonVision Functionality

### LINKED ISSUE #141: Use photonvision for turret control

riversry:
## Justification

Limelight sucks, photonvision will be better

## Implementaion

Created a VisionSubsystem interface to allow for interchangeable Limelight and PhotonVision use
Created PhotonSubsystem to enable PhotonVision use
Changed turret control to use VisionSubsystem interface

## Testing Done

No testing

Fixes #141


src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -184,24 +185,26 @@ public Command testTurretCommand(double degrees) {
...
-
-      if (Math.abs(Math.toRadians(limelightTX) + turretPosition)
-          > Math.toRadians(Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES)) {
-        // TODO: Make turret rotate the drivebase if necessary and driver thinks it's a good idea
+    if (visionSubsystem != null) {
+      int tagID =
+          ((robotStateManager.getAllianceColor() == AllianceColor.BLUE)
+              ? Constants.TurretConstants.SPEAKER_TAG_ID_BLUE
+              : Constants.TurretConstants.SPEAKER_TAG_ID_RED);
+      double limelightTX = visionSubsystem.getTurretYaw(tagID);
```
lohengrin332, line 193: Since you're converting to using an interface instead of using limelight directly, I'd recommend calling things `vision*` instead of `limelight*`.

src/main/java/frc/robot/subsystems/TurretSubsystem.java
```diff
@@ -184,24 +185,26 @@ public Command testTurretCommand(double degrees) {
...
+        setTurretPos(Math.toRadians(limelightTX) + turretPosition);
+
+        // Y & Tilting
+        double limelightTY = visionSubsystem.getTurretPitch(tagID);
+        double distanceToTag = tyToDistanceFromTag(limelightTY);
+        tagDistanceEntry.log(distanceToTag);
+        // TODO: Add vertical tilt and use distance for it
+
+        if (Math.abs(Math.toRadians(limelightTX) + turretPosition)
+            > Math.toRadians(Constants.TurretConstants.MAX_TURRET_ANGLE_DEGREES)) {
```
lohengrin332, line 205: As I understand it, starting from pointing straight forward, the turret can go 110 degrees one direction, 5 degrees in the other (or something like that). Is that correct? This code assumes that the turret can go 110 degrees in either direction.

src/main/java/frc/robot/subsystems/vision/VisionSubsystem.java
```diff
@@ -0,0 +1,16 @@
+package frc.robot.subsystems.vision;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Pose3d;
+
+public interface VisionSubsystem {
+  public Pose3d getPose3d();
```
lohengrin332, line 7: These first 3 methods aren't used outside of their implementing classes, so they probably don't need to be part of the interface.

I also think that once you implement the refactor to remove the excess calls to `postEstimator.update()`, you will need to make these private for them to make any sense.

src/main/java/frc/robot/subsystems/vision/VisionSubsystem.java
```diff
@@ -0,0 +1,16 @@
+package frc.robot.subsystems.vision;
+
+import edu.wpi.first.math.geometry.Pose2d;
+import edu.wpi.first.math.geometry.Pose3d;
+
+public interface VisionSubsystem {
```
lohengrin332, line 6: I think that you will always want a `VisionSubsystem` to extend `SubsystemBase`. If so, you might consider making this an abstract class instead of an interface, and extend `SubsystemBase` here. E.g.:

```suggestion
public abstract class VisionSubsystem extends SubsystemBase {
```

You would then need to make the method definitions `abstract`, e.g. `public abstract double getTurretYaw(int id);`.

Not a huge change and not required, but I think it simplifies things because you can use `VisionSubsystem` as a variable type, instead of having to switch things out depending on which one has been instantiated.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -103,15 +100,31 @@ public RobotContainer() {
...
       trapElvSubsystem = new TrapElvSubsystem();
     } else {
       trapElvSubsystem = null;
     }
+    if (Constants.enabledSubsystems.turretEnabled) {
+      if (Constants.enabledSubsystems.limeLightEnabled) {
+        turretSubsystem = new TurretSubsystem(robotStateManager, limelightSubsystem);
+      } else if (Constants.enabledSubsystems.photonEnabled) {
+        turretSubsystem = new TurretSubsystem(robotStateManager, photonSubsystem);
+      } else {
```
caseyjbrotherton, line 124: We have two vision systems on the robot.  One for positioning, and the other for targeting.  Do we need a second photonSubsystem/limelightSubsystem here?

src/main/java/frc/robot/subsystems/TurretSubsystem.java
caseyjbrotherton: commit 329845d has the changes for an asymmetric turret rotation.  

Looks like that is in TurretVerticalControl branch.  Doesn't look like it is in main, yet

src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
```diff
@@ -0,0 +1,153 @@
...
+      new DebugEntry<Double>(0.0, "Distance To Tag 3 (m)", this);
+  private DebugEntry<Double> distanceEntryTag4 =
+      new DebugEntry<Double>(0.0, "Distance To Tag 4 (m)", this);
+
+  public PhotonSubsystem(BiConsumer<Pose2d, Double> measurementConsumer) {
+    this.measurementConsumer = measurementConsumer;
+    camera = new PhotonCamera("Camera_Module_v1");
+    while (!camera.isConnected()) {
+      camera = new PhotonCamera("Camera_Module_v1");
+    }
```
caseyjbrotherton, line 51: Think we wanted to remove this?  

Maybe try a limited number of times, if we still are concerned, ( like three times ) and log a driverstation error for each loop when it isn't connected.

src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
```diff
@@ -0,0 +1,153 @@
...
+    result = camera.getLatestResult();
+    aprilTagFieldLayout = AprilTagFields.k2024Crescendo.loadAprilTagLayoutField();
+    poseEstimator =
+        new PhotonPoseEstimator(
+            aprilTagFieldLayout,
+            PoseStrategy.MULTI_TAG_PNP_ON_COPROCESSOR,
+            camera,
+            new Transform3d( // camera offsets will go here when we have them finalized
+                new Translation3d(0, 0, 0),
```
caseyjbrotherton, line 60: Don't we have camera offsets in constants?  At least the height and pitch.  

Let's get the translation in constants so we have one place to adjust.

I was looking for the limelight translation for the camera, and I didn't see it.  was it set only in the limelight UI?

Let's add that to the Limelightsubsystem as well, so that everything is in code

src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
```diff
@@ -0,0 +1,153 @@
...
+
+  public void periodic() {
+    if (Robot.isReal()) {
+      result = camera.getLatestResult();
+      resultLatency = result.getLatencyMillis();
+      if (result.hasTargets()) {
+        targets = result.getTargets();
+        if (targets.size() > 1) {
+          measurementsUsed++;
+          measurementConsumer.accept(getPose2d(), getTime() - (resultLatency / 1000));
```
caseyjbrotherton, line 125: https://docs.photonvision.org/en/latest/docs/programming/photonlib/robot-pose-estimator.html#using-a-photonposeestimator

Even better.  PhotonVision tracks the time when the pose was captured.  So, you don't need to keep the resultLatency.
Just use getTime

src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
```diff
@@ -0,0 +1,153 @@
...
+          if (measurementsUsed % 100 == 0) {
+            measurementEntry.log(measurementsUsed);
+          }
+        }
+        // logging stuff
+        for (PhotonTrackedTarget target : targets) {
```
caseyjbrotherton, line 131: I still like these calculations, but wondering if we should put them inside of an isCompetiton block.

THoughts?

src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
Malachi-lewis: I created constants for the alphabot's limelight mount, but we can change them after we get the values for the actual limelight mount

Malachi-lewis: Somehow I became malachi, this was ryan

src/main/java/frc/robot/subsystems/TurretSubsystem.java
riversry: I changed them, it was just a leftover piece of code from the old version using only limelight

src/main/java/frc/robot/RobotContainer.java
riversry: I have compatibility for two photonvision cameras, but if we're planning to run both limelight and photonvision then I'll need to change more things around.

src/main/java/frc/robot/subsystems/vision/VisionSubsystem.java
riversry: I made a card in the backlog for it, and we can discuss it more on Saturday because I'm not entirely sure how abstract classes work

lohengrin332: It's pretty easy; I think if we sat down together we could convert it in about 10 minutes.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -97,14 +97,19 @@ public RobotContainer() {
...
     }
-    triggerSubsystem = new TriggerSubsystem();
-    if (Constants.enabledSubsystems.limeLightEnabled
-        && Constants.enabledSubsystems.drivetrainEnabled) {
-      visionSubsystem =
-          new LimelightSubsystem(drivetrain.getVisionMeasurementConsumer(), robotStateManager);
-    } else if (Constants.enabledSubsystems.photonEnabled
-        && Constants.enabledSubsystems.drivetrainEnabled) {
-      visionSubsystem = new PhotonSubsystem(drivetrain.getVisionMeasurementConsumer());
+    if (Constants.enabledSubsystems.triggerEnabled) {
```
lohengrin332, line 100: Not a required change, but I like to use ternaries for this type of thing.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -196,7 +202,7 @@ private void configureBindings() {
     // Shooter commands
     if (Constants.enabledSubsystems.shooterEnabled
```
lohengrin332, line 215: This if/else block is probably more readable as:

```java
if (shooterEnabled && triggerEnabled) {
  if (visionEnabled) {
    ...
  } else {
    ...
  }
}
```

src/main/java/frc/robot/RobotContainer.java
riversry: It should be resolved in the new commits

src/main/java/frc/robot/RobotContainer.java
lohengrin332: Making this change would also allow you to extract `shooterSubsystem.setDefaultCommand(shooterSubsystem.shooterIdle());` to the outer if, and only write it once:

```java
if (shooterEnabled && triggerEnabled) {
  shooterSubsystem.setDefaultCommand(shooterSubsystem.shooterIdle());
  if (visionEnabled) {
    ...
  } else {
    ...
  }
}
```

src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
```diff
@@ -22,9 +22,9 @@
 import org.photonvision.targeting.PhotonPipelineResult;
 import org.photonvision.targeting.PhotonTrackedTarget;
 
-public class PhotonSubsystem extends SubsystemBase implements VisionSubsystem {
-  private int measurementsUsed = 0;
-  private DebugEntry<Integer> measurementEntry = new DebugEntry<Integer>(0, "measurements", this);
+public class PhotonSubsystem extends VisionSubsystem {
+  private double measurementsUsed = 0;
```
lohengrin332, line 26: Why was this changed to a double? Iterators are usually ints. In some languages float types _can't_ be iterators.

src/main/java/frc/robot/RobotContainer.java
riversry: Done

src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
riversry: The DebugEntry class doesn't have functionality to accept Integers so I replaced it with a Double, it should work the same I believe

lohengrin332 (commented):
I have a few more questions/suggestions. We can work through this together if that would work better for you.

lohengrin332 (approved):

lohengrin332 (approved):

PR #169, opened 2024-02-23
Tof for wrist

Malachi-lewis:
## Justification
We will have Time of Flight Sensors on the final robot. Hence, we need code for them. We also need them soon so that we can safely intake pieces without prematurely advancing to the next step.

## Implementaion
I created a static class with BooleanSuppliers and Triggers to get the states of the sensors.

## Testing Done
No.


src/main/java/frc/robot/Constants.java
```diff
@@ -183,19 +199,30 @@ public static class TrapElvConstants {
...
-    public static final double DRUM_RADIUS = Units.inchesToMeters(1);
     public static final double ELV_MIN_HEIGHT = Units.inchesToMeters(12);
     public static final double ELV_MAX_HEIGHT = Units.inchesToMeters(30);
+    public static final double DRUM_RADIUS = Units.inchesToMeters(1);
+
+    public static final double WRIST_MIN_ANGLE = Units.degreesToRadians(-90); // RADS
+    public static final double WRIST_MAX_ANGLE = Units.degreesToRadians(270); // RADS
+    public static final double WRIST_LENGTH = Units.inchesToMeters(11.877934);
+    public static final double WRIST_MOI = 0.3175242664; // Moment of Inertia
+    public static final double WRIST_GEAR_RATIO = 32.727272727272727272;
```
lohengrin332, line 216: I think this should be written as `360 / 11`, since that is more understandable number; there are 360 degrees over 11 teeth.

lohengrin332: Also, I believe Java does the math at compile time, so it's ultimately just as efficient to use the fraction.

src/main/java/frc/robot/utilities/TOFSensorSimple.java
```diff
@@ -7,23 +7,45 @@
...
+
+  public void updateSimBool(boolean isBeamBroke) {
+    isBeamBrokeSim = isBeamBroke;
   }
 
   public double getMilliMeters() {
     return sensor.getRange();
   }
 
   public boolean isBeamBroke() {
```
lohengrin332, line 37: Probably should be `isBeamBroken`...

src/main/java/frc/robot/utilities/TOFSensorSimple.java
```diff
@@ -7,23 +7,45 @@
...
-    return getMilliMeters() < threshold;
+    if (Robot.isReal()) {
+      return getMilliMeters() < threshold;
+    }
+    return isBeamBrokeSim;
+  }
+
+  public boolean isBeamBrokeInverse() {
```
lohengrin332, line 44: Which would make this `isBeamUnbroken`.

lohengrin332: Also it looks like this method is never used; instead we're using `!isBeamBroke()`, so let's get rid of this one.

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -134,22 +136,25 @@ public TrapElvSubsystem() {
...
-    sourceBreak = new DigitalInput(TrapElvConstants.SOURCE_BREAK_ID);
-    groundBreak = new DigitalInput(TrapElvConstants.GROUND_BREAK_ID);
-
     // Elv
     if (isElv) {
-      baseLimit = new DigitalInput(TrapElvConstants.BASE_BREAK_ID);
-      scoringLimit = new DigitalInput(TrapElvConstants.SCORING_BREAK_ID);
-      baseLog = new DebugEntry<Boolean>(baseLimit.get(), "Base Limit Switch", this);
-      scoringLog = new DebugEntry<Boolean>(scoringLimit.get(), "Scoring Limit Switch", this);
+      // baseLimit = new DigitalInput(TrapElvConstants.BASE_BREAK_ID);
```
lohengrin332, line 154: Please remove commented code.

lohengrin332 (changes requested):
Several changes have been requested in comments. Please address.

src/main/java/frc/robot/utilities/TOFSensorSimple.java
Malachi-lewis: fixed

src/main/java/frc/robot/Constants.java
Malachi-lewis: done

lohengrin332 (approved):

src/main/java/frc/robot/utilities/TOFSensorSimple.java
```diff
@@ -7,27 +7,42 @@
...
 
   public TOFSensorSimple(int ID, double threshold) {
     sensor = new TimeOfFlight(ID);
     this.threshold = threshold; // in mm
+    isBeamBrokeSim = false;
+  }
+
+  public TOFSensorSimple(int ID) {
+    sensor = new TimeOfFlight(ID);
+    this.threshold = 1; // in mm
```
caseyjbrotherton, line 26: 1mm is a really small distance.  Should we change this default to something more likely to be usable?

src/main/java/frc/robot/utilities/TOFSensorSimple.java
```diff
@@ -7,27 +7,42 @@
...
+import frc.robot.Robot;
 
 /** Add your docs here. */
 public class TOFSensorSimple {
   private TimeOfFlight sensor;
   private double threshold;
+  private boolean isBeamBrokeSim;
 
   public TOFSensorSimple(int ID, double threshold) {
     sensor = new TimeOfFlight(ID);
```
caseyjbrotherton, line 19: We should explicitly set the ranging mode to Short:
https://www.playingwithfusion.com/frc/2022/javadoc/com/playingwithfusion/TimeOfFlight.html#setRangingMode(com.playingwithfusion.TimeOfFlight.RangingMode,double)

This should be the default, but we want to make sure as it has better readings in bright light with a tradeoff of only 1.3 meters.  We can also explicitly set the time to as low as 24ms.

( Doesn't mention the default. )

src/main/java/frc/robot/Constants.java
```diff
@@ -155,12 +155,28 @@ public static class TurretConstants {
...
   }
 
+  public static class SensorConstants {
+    // TODO: Get real ids
+    public static final int SOURCE_BREAK_ID = 1;
+    public static final int GROUND_BREAK_ID = 2;
+    public static final int BASE_BREAK_ID = 3;
+    public static final int SCORING_BREAK_ID = 4;
+
+    public static final double WRIST_BREAK_THOLD = 0.1;
```
caseyjbrotherton, line 177: .1 mm is too small.  The note is 12 inches wide.  If we could the very peak of the torus, that would be about 6 inches away?  Can we start with that as a guess?

src/main/java/frc/robot/Constants.java
```diff
@@ -155,12 +155,28 @@ public static class TurretConstants {
...
+    public static final double TURRET_RADIUS = Units.inchesToMeters(12);
   }
 
   public static class OperatorConstants {
     public static final int kDriverControllerPort = 0;
   }
 
+  public static class SensorConstants {
+    // TODO: Get real ids
+    public static final int SOURCE_BREAK_ID = 1;
```
caseyjbrotherton, line 172: I think there are one too many sensors here.

We have a plan for one sensor on the intake.
One sensor between the trigger rollers and the shooter rollers

And one sensor on the amp arm.

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -234,9 +238,11 @@ private void configureBindings() {
     // Trap Elv Intaking
     if (Constants.enabledSubsystems.elvEnabled) {
       OI.getButton(OI.Driver.groundIntakeButton)
-          .whileTrue(trapElvSubsystem.intakeGround().onlyWhile(trapElvSubsystem.getGroundBreak()));
+          .whileTrue(
+              trapElvSubsystem.intakeGround().onlyWhile(SensorManager.getGroundBreakBoolInverse()));
       OI.getButton(OI.Driver.sourceIntakeButton)
-          .whileTrue(trapElvSubsystem.intakeSource().onlyWhile(trapElvSubsystem.getSourceBreak()));
+          .whileTrue(
+              trapElvSubsystem.intakeSource().onlyWhile(SensorManager.getSourceBreakBoolInverse()));
```
caseyjbrotherton, line 245: I think that sourcebreak and groundbreak are actually the same sensor.

If I understand the naming.

caseyjbrotherton (changes requested):

PR #171, opened 2024-02-23
Feature/climber

Malachi-lewis:
## Justification
We need a climber

## Testing Done
None yet.


src/main/java/frc/robot/subsystems/ClimberSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+    public MinimumDelay(double time, BooleanSupplier base) {
+      this.base = base;
+      this.wait = time;
+      timer = new Timer();
+    }
+
+    @Override
+    public boolean getAsBoolean() {
+      timer.start();
+      if (timer.advanceIfElapsed(wait)) {
```
azvanderpas, line 96: I think you want to use hasElapsed() instead of advanceIfElapsed(). 

With advanceIfElapsed(), the `base` is only checked once every `wait` interval. So if `base` is the button press, it would have to be held down (or have an operator with an inhumanly accurate internal clock).

src/main/java/frc/robot/subsystems/ClimberSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  public Command climberSequence(BooleanSupplier step) {
+    return Commands.sequence(
+        gotoPositionCommand(ClimberConstants.IDLE_SETPOINT),
+        Commands.waitUntil(step),
+        gotoPositionCommand(ClimberConstants.PICK_UP),
+        Commands.waitUntil(step),
+        Commands.startEnd(() -> applyDemand(ClimberConstants.CLIP_VOLTAGE), () -> {}, this)
+            .until(new MinimumDelay(ClimberConstants.MINIMUM_WAIT, step)),
```
azvanderpas, line 78: Command compositions allows this functionality without having to write a special class.

```suggestion
        Commands.startEnd(() -> applyDemand(ClimberConstants.CLIP_VOLTAGE), () -> {}, this)
            .raceWith(
                Commands.waitSeconds(ClimberConstants.MINIMUM_WAIT)
                    .andThen(Commands.waitUntil(step))),
```

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -52,6 +53,7 @@ public class RobotContainer {
   private final ShooterSubsystem shooterSubsystem;
   private final TriggerSubsystem triggerSubsystem;
   private final TurretSubsystem turretSubsystem;
+  private final ClimberSubsystem climberSubsystem;
```
azvanderpas, line 56: Need to add some button mappings

src/main/java/frc/robot/Constants.java
```diff
@@ -163,4 +164,24 @@ public static class enabledSubsystems {
 
   // Lights
   public static final int LED_COUNT = 20;
+
+  public static class ClimberConstants {
+    public static final double GEAR_RATIO = 0;
+    public static final double MAX_ERROR = 0;
+    public static final double PICK_UP = 0;
+    public static final double CLIP_VOLTAGE = 0;
```
azvanderpas, line 172: What does CLIP_VOLTAGE refer to?

src/main/java/frc/robot/subsystems/ClimberSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+    leftEncoder = leftMotorController.getEncoder();
+    rightEncoder = rightMotorController.getEncoder();
+
+    leftPidController.setP(ClimberConstants.POSITION_P_GAIN);
+    leftPidController.setI(ClimberConstants.POSITION_I_GAIN);
+    leftPidController.setD(ClimberConstants.POSITION_D_GAIN);
+
+    rightPidController.setP(ClimberConstants.POSITION_P_GAIN);
+    rightPidController.setI(ClimberConstants.POSITION_I_GAIN);
+    rightPidController.setD(ClimberConstants.POSITION_D_GAIN);
```
azvanderpas, line 48: Going to want to be able to change these in runtime. Add some shuffleboard entries?

src/main/java/frc/robot/subsystems/ClimberSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+    target = rotation * Constants.ClimberConstants.GEAR_RATIO;
+    leftPidController.setReference(target, ControlType.kPosition);
+    rightPidController.setReference(target, ControlType.kPosition);
+  }
+
+  public Command gotoPositionCommand(double position) {
+    Runnable init = () -> gotoPosition(position);
+    BooleanSupplier done =
+        () ->
+            rightEncoder.getPosition() - target < Constants.ClimberConstants.MAX_ERROR
```
caseyjbrotherton, line 65: absolute value for the max error?

src/main/java/frc/robot/subsystems/ClimberSubsystem.java
```diff
@@ -0,0 +1,102 @@
...
+    rightMotorController.restoreFactoryDefaults();
+
+    leftMotorController.setInverted(ClimberConstants.LEFT_IS_INVERTED);
+    rightMotorController.setInverted(ClimberConstants.RIGHT_IS_INVERTED);
+
+    leftPidController = leftMotorController.getPIDController();
+    rightPidController = rightMotorController.getPIDController();
+
+    leftEncoder = leftMotorController.getEncoder();
+    rightEncoder = rightMotorController.getEncoder();
```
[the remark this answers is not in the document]
Shazamboo: nope

src/main/java/frc/robot/subsystems/ClimberSubsystem.java
Shazamboo: this command is no longer being used.

src/main/java/frc/robot/subsystems/ClimberSubsystem.java
Shazamboo: I added an absolute value method around the pos - target so that we can always check against the constant max error value.

src/main/java/frc/robot/Constants.java
Shazamboo: the constant voltage used while the climber arm is lowering onto the chain. We are using a constant voltage instead of a set position for this specific step in the climb process because the height of the chain varies depending on where we try to climb from, so setting a specific position will not work well.

src/main/java/frc/robot/RobotContainer.java
Shazamboo: Kian is working on that in a separate branch.

src/main/java/frc/robot/subsystems/ClimberSubsystem.java
```diff
@@ -0,0 +1,300 @@
...
+        new DebugEntry<Double>(rightMotorController.get(), "Right Climber Output", this);
+  }
+
+  public void applyDemand() {
+    leftMotorController.setVoltage(ClimberConstants.CLIP_VOLTAGE);
+    rightMotorController.setVoltage(ClimberConstants.CLIP_VOLTAGE);
+  }
+
+  public void stopMotors() {
+    leftMotorController.setVoltage(0);
```
caseyjbrotherton, line 182: Is this change for deploying or for simulation, or both?

caseyjbrotherton: We could do an "isSimulation" if statement here if we want stopMotor for the real

src/main/java/frc/robot/subsystems/ClimberSubsystem.java
```diff
@@ -0,0 +1,300 @@
...
+      } else {
+        leftMotorController.set(0);
+        rightMotorController.set(0);
+      }
+    } else {
+      leftPidController.setReference(state.getMotorAngle(), ControlType.kPosition);
+      rightPidController.setReference(state.getMotorAngle(), ControlType.kPosition);
+    }
+
+    // Runnable init = () -> gotoPosition(state.getStateAngle());
```
caseyjbrotherton, line 219: let's remove this commented code

src/main/java/frc/robot/subsystems/TrapElvSubsystem.java
```diff
@@ -40,15 +40,15 @@
 import java.util.function.BooleanSupplier;
 
 public class TrapElvSubsystem extends SubsystemBase {
-  private final boolean isElv = false;
+  private boolean isElv = false;
```
caseyjbrotherton, line 43: These finals were removed for testing?  Or is there a required reason that these are no longer finals?

PR #172, opened 2024-02-24
Auton Paths

riversry:
## Justification

Auton is nice to have 👍

## Implementaion

The majority of work done in this branch should be focused on pathplanner, with few code changes
Most paths have been written with the expectation of a fully functional turret, but I did write some paths specifically for Waco while our turret isn't at full range of motion.

## Testing Done

Very little testing done so far, and future testing depends on how much robot time I get later on.


src/main/java/frc/robot/config/TunerConstants.java
```diff
@@ -43,7 +43,7 @@ public class TunerConstants {
 
   private static final double kDriveGearRatio = 6.122448979591837;
   private static final double kSteerGearRatio = 21.428571428571427;
-  private static final double kWheelRadiusInches = 2;
+  private static final double kWheelRadiusInches = 1.94; // 1.924; // 1.89575; - old values
```
azvanderpas, line 46: Remove comments of old values

src/main/java/frc/robot/config/TunerConstants.java
```diff
@@ -86,32 +86,32 @@ public class TunerConstants {
   private static final int kFrontLeftSteerMotorId = 2;
   private static final int kFrontLeftEncoderId = 1;
 
-  private static final double kFrontLeftXPosInches = 12;
-  private static final double kFrontLeftYPosInches = 12;
+  private static final double kFrontLeftXPosInches = 12.375;
```
azvanderpas, line 89: This puts the wheels outside the robot perimeter, doesn't it? 

0.61 meters (from pathplanner settings.json) is 24.0157 inches. 
12.375 inches * 2 = 24.75 inches.

azvanderpas (changes requested):

src/main/java/frc/robot/config/TunerConstants.java
Malachi-lewis: We'll be leaving them as 12 for waco but we can measure and adjust them after

Malachi-lewis: i became malachi again, this was ryan

azvanderpas:
Will get merged with #179

PR #175, opened 2024-02-24
Stationary Turret Shooting

riversry:
## Justification

Turret will not be turning or using pitch control for waco

## Implementaion

This button binding should allow us to aim using the drivebase instead of the turret

## Testing Done

No testing has been done so far because the robot is being worked on


KianChilds (approved):

caseyjbrotherton (approved):
Tested on simulation

PR #181, opened 2024-02-28
Comp Shuffleboard

ModBoyex:
## Justification
Too many things on network tables will slow the robot down.

## Implementaion
turn anything that is not needed during the match to a debug log so that they are logged for looking after the game, but are not clogging the board during the match.


src/main/java/frc/robot/subsystems/trapElvSubsystem/TrapElvSubsystem.java
```diff
@@ -82,16 +82,18 @@ public class TrapElvSubsystem extends SubsystemBase {
...
+  private DebugEntry<String> currentWristStateEntry;
 
-  private ShuffleboardTab TrapElvTab = Shuffleboard.getTab("TrapElvSubsystem");
-  private GenericEntry FFOutput = TrapElvTab.add("FF Output", 0).withPosition(8, 0).getEntry();
-  private GenericEntry wristOutput =
-      TrapElvTab.add("Wrist Motor Output", 0).withPosition(7, 0).getEntry();
-  private GenericEntry wristGoal = TrapElvTab.add("Wrist Goal", 0).withPosition(6, 0).getEntry();
-  private GenericEntry currentWristStateEntry =
-      TrapElvTab.add("Current Wrist State", TrapElvState.STOWED.name()).getEntry();
+  private TunableNumber wristP;
```
caseyjbrotherton, line 94: Should we just lock these numbers in, instead of having tunable numbers?  Thoughts?

ModBoyex: Could be nice to have is things get weird at comp for some reason

PR #201, opened 2024-03-07
PID Value Class

ModBoyex:
## Justification
Having to make 4 PID Constants, 4 Tunnable numbers, and 4 more lines of code to set each value, gets annoying anytime you have to make a new PID.

## Implementaion
I created a class that deals with setting all four vars in one line, making a tunable number for every value, and giving back a PID controller that has all the values set.

### DIFF
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -261,7 +261,7 @@ public final class Constants {
     public static final boolean signalEnabled = false;
     public static final boolean shooterEnabled = true;
     public static final boolean triggerEnabled = true;
-    public static final boolean turretEnabled = true;
+    public static final boolean turretEnabled = false;
     public static final boolean climberEnabled = true;
   }
 
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -7,6 +7,8 @@ package frc.robot;
 import edu.wpi.first.math.geometry.Translation2d;
 import edu.wpi.first.math.util.Units;
 import frc.robot.subsystems.shooterSubsystem.ShooterSubsystem.SpeakerConfig;
+import frc.robot.utilities.HowdyFF;
+import frc.robot.utilities.HowdyPID;
 
 /**
  * The Constants class provides a convenient place for teams to hold robot-wide numerical or boolean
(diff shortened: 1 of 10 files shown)


src/main/java/frc/robot/utilities/HowdyPID.java
```diff
@@ -0,0 +1,117 @@
...
+    motor.getPIDController().setFF(this.FF);
+  }
+
+  public void setSparkPidController(CANSparkMaxSim motor) {
+    motor.getPIDController().setP(this.P);
+    motor.getPIDController().setI(this.I);
+    motor.getPIDController().setD(this.D);
+    motor.getPIDController().setIZone(this.Iz);
+    motor.getPIDController().setFF(this.FF);
+  }
```
azvanderpas, line 90: Since CANSparkMaxSim extends CANSparkMax, this method isn't needed

src/main/java/frc/robot/utilities/HowdyPID.java
```diff
@@ -0,0 +1,117 @@
...
+
+  public void getSparkPidController(CANSparkMax motor) {
+    motor.getPIDController().setP(this.P);
+    motor.getPIDController().setI(this.I);
+    motor.getPIDController().setD(this.D);
+    motor.getPIDController().setIZone(this.Iz);
+    motor.getPIDController().setFF(this.FF);
+  }
+
+  public void setSparkPidController(CANSparkMaxSim motor) {
```
azvanderpas, line 84:
```suggestion
  public void setSparkPidController(CANSparkMax motor) {
```

src/main/java/frc/robot/utilities/HowdyPID.java
```diff
@@ -0,0 +1,117 @@
...
+
+  public void createTunableNumbers(String name, PIDController controller, Subsystem subsystem) {
+    this.tuneP = new TunableNumber(name.concat(" P"), this.P, p -> controller.setP(p), subsystem);
+    this.tuneI = new TunableNumber(name.concat(" I"), this.I, i -> controller.setI(i), subsystem);
+    this.tuneD = new TunableNumber(name.concat(" D"), this.D, d -> controller.setD(d), subsystem);
+    this.tuneIz =
+        new TunableNumber(name.concat(" Iz"), this.Iz, iz -> controller.setIZone(iz), subsystem);
+  }
+
+  public void getSparkPidController(CANSparkMax motor) {
```
azvanderpas, line 76: set not get?

src/main/java/frc/robot/utilities/HowdyFF.java
```diff
@@ -0,0 +1,49 @@
...
+
+  public double getKV() {
+    return this.kV;
+  }
+
+  public double getKA() {
+    return this.kA;
+  }
+
+  public ArmFeedforward getArmFeedforward() {
```
azvanderpas, line 46: Add other feedforward types? Elevator and SimpleMotor

src/main/java/frc/robot/Constants.java
```diff
@@ -283,7 +261,7 @@ public static class enabledSubsystems {
     public static final boolean signalEnabled = false;
     public static final boolean shooterEnabled = true;
     public static final boolean triggerEnabled = true;
-    public static final boolean turretEnabled = false;
+    public static final boolean turretEnabled = true;
```
azvanderpas, line 264: Keep false for now

src/main/java/frc/robot/utilities/HowdyFF.java
ModBoyex: we are only using ArmFF for the moment, but i guess this would be nice for future proofing the class

azvanderpas (approved):

PR #164, opened 2024-02-19
Make unit tests

riversry:
## Justification

Made small fixes to PhotonSubsystem to ensure checkPoseValidity method works as intended through JUnit testing
Changed some numbers to be constants

## Testing Done

Fixes #157 
Fixes #160

PR #202, opened 2024-03-07
PhotonVision Updates

riversry:
## Justification

We're planning to have both the turret and main cameras on the robot for San Antonio, so I implemented better control for using photonVision
## Implementaion

Created a constant to disable the turret camera
Edited the 2d tag estimation to allow for turret camera disabled

I also need to reflash the main camera that's currently on the robot and upload the settings and calibration files
## Testing Done

None so far, robot was occupied

### DIFF
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -142,7 +142,11 @@ public class RobotContainer {
     }
     trapElvCommandFactory = new TrapElvCommandFactory(trapElvSubsystem);
     if (enabledSubsystems.turretEnabled) {
-      turretSubsystem = new TurretSubsystem(robotStateManager, null);
+      if (visionSubsystem != null) {
+        turretSubsystem = new TurretSubsystem(robotStateManager, visionSubsystem);
+      } else {
+        turretSubsystem = new TurretSubsystem(robotStateManager, null);
+      }
     } else {
       turretSubsystem = null;
     }
--- a/src/main/java/frc/robot/RobotContainer.java
+++ b/src/main/java/frc/robot/RobotContainer.java
@@ -252,7 +252,8 @@ public class RobotContainer {
             Commands.startEnd(
                 () -> OI.Driver.setRumble(Constants.OperatorConstants.RUMBLE_STRENGTH),
                 () -> OI.Driver.setRumble(0)));
-    new Trigger(shooterSubsystem::isShooterReady)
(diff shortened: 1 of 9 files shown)


src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
```diff
@@ -25,6 +25,9 @@
 import org.photonvision.targeting.PhotonTrackedTarget;
 
 public class PhotonSubsystem extends SubsystemBase implements VisionSubsystem {
+
+  private final boolean TURRET_CAMERA_ENABLED = false;
```
caseyjbrotherton, line 29: Camera is mounted, let's set this to True

riversry: I was waiting until the second camera was physically mounted because the current camera is the "main" camera.

src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
```diff
@@ -118,73 +126,101 @@ private boolean checkPoseValidity(EstimatedRobotPose pose) {
...
+  public void periodic() {
+    if (!Robot.isReal()) return;
+
+    mainResult = mainCamera.getLatestResult();
+    if (!mainResult.hasTargets()) return;
+
+    List<PhotonTrackedTarget> targets = mainResult.getTargets();
+    if (targets.size() > 1) {
+      EstimatedRobotPose newPose = getPVEstimatedPose();
+      if ((newPose.estimatedPose.getX() > 12) || (newPose.estimatedPose.getX() < 4.54)) {
```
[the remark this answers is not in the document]
riversry: This was fixed in pull request #164 but was never merged

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -142,7 +142,11 @@ public RobotContainer() {
     }
     trapElvCommandFactory = new TrapElvCommandFactory(trapElvSubsystem);
     if (enabledSubsystems.turretEnabled) {
-      turretSubsystem = new TurretSubsystem(robotStateManager, null);
+      if (visionSubsystem != null) {
```
azvanderpas, line 145: The visionSubsystem disable got changed such that it is never null, but it could be a default interface class. Double check this will work the way you expect, or potentially change VisionSubsystem back to an abstract class from an interface.

PR #204, opened 2024-03-09
Fix climb encoder

Josh-Westbrook:
## Justification
Fix relative encoder so it knows the actual difference of arm angle


src/main/java/frc/robot/subsystems/climberSubsystem/ClimberSubsystem.java
```diff
@@ -53,18 +53,15 @@ private void configMotor(CANSparkMax motor, boolean invert) {
     double armPosition;
 
     armPosition = motor.getAbsoluteEncoder().getPosition();
-
+    if(armPosition>0.5) {
+      armPosition = armPosition-1;
+    }
```
ModBoyex, line 58: I would change this to be:
`if(armPosition>0.5) armPosition -= 1`
its just a bit cleaner in my opinion

azvanderpas (approved):

PR #212, opened 2024-03-10
ShooterPostWaco

aetherstorm-dev:
Some post-Waco shooter updates. Mostly for convenience.


src/main/java/frc/robot/Constants.java
```diff
@@ -58,10 +58,10 @@ public static class ShooterConstants {
     public static final double SHOOTER_LEFT_D = 0;
     public static final double SHOOTER_LEFT_FF = 0;
 
-    public static final double SHOOTER_P = 0.001;
-    public static final double SHOOTER_I = 0.00000035;
-    public static final double SHOOTER_D = 0;
-    public static final double SHOOTER_FF = 0;
+    public static final double SHOOTER_RIGHT_P = 0.001;
```
caseyjbrotherton, line 61: You will want to merge from main.  There is a new HowdyPID object that is defined for shooter's pid

src/test/java/TestShooterSubsystem.java
```diff
@@ -1,43 +1,18 @@
...
-      speedsPair = ShooterSubsystem.calculateShooterSpeeds(speakerConfigListTest[i][0]);
-      speedsArray[0] = speedsPair.getSpeedLeftInRPM();
-      speedsArray[1] = speedsPair.getSpeedRightInRPM();
-      configSpeeds = arraySlice(speakerConfigListTest[i], 1, 2);
-      Assertions.assertArrayEquals(configSpeeds, speedsArray);
+    for (int i = 0; i < ShooterSubsystem.speakerConfigList.length; i++) {
+      testConfig =
+          ShooterSubsystem.calculateShooterSpeeds(
+              ShooterSubsystem.speakerConfigList[i].getDistanceInInches());
+      Assertions.assertEquals(testConfig, ShooterSubsystem.speakerConfigList[i]);
```
caseyjbrotherton, line 15: This test only tests exact matches for Shooter Configurations.

We should also consider test cases:

- Is config list in sorted order, with both distance, left speed and right speed increasing as we move through the list?
- Does a distance value below minimum provide the lowest possible speeds?
- Does a distance value above maximum provide the highest possible speed?

caseyjbrotherton (changes requested):
Code looks good.  Much cleaner with equals.

Since we are changing, let's consider adding more test cases to verify operation

src/main/java/frc/robot/Constants.java
aetherstorm-dev: Resolved in bb39de9

src/test/java/TestShooterSubsystem.java
aetherstorm-dev: Tests added in 3cc5cbb

PR #213, opened 2024-03-10
re add wirst scoring auto command + TOF Tweaks

ModBoyex:
## Justification

The command for scoring Amp in auto was commented out. this pull request re-implements it and adds the TOF as a factor in running the rollers.

## Testing Done

This still needs to be tested, but might be able to be tested at Westlake


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -309,18 +308,23 @@ private Command shootAuton() {
   }
 
   private Command ampAuton() {
-    return null; // return Commands.parallel(
-    //     trapElvCommandFactory.positionAMP(),
-    //     Commands.waitUntil(trapElvSubsystem.isAMPReady())
-    //         .andThen(trapElvCommandFactory.scoreAMP()));
+    return Commands.parallel(
+        trapElvCommandFactory.positionAMP(),
+        Commands.waitUntil(trapElvSubsystem.isAMPReady())
```
azvanderpas, line 313: If this subsystem is disabled, this will cause a crash

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -309,18 +308,23 @@ private Command shootAuton() {
...
+            .onlyWhile(trapElvCommandFactory.getSourceBreak()));
   }
 
   // Register commands for auton
   public void registerCommands() {
     HashMap<String, Command> autonCommands = new HashMap<String, Command>();
 
-    autonCommands.put("Shoot", shootAuton());
-    // autonCommands.put("Amp", ampAuton());
+    if (enabledSubsystems.shooterEnabled) {
```
azvanderpas, line 322: Shouldn't need this if we setup our command factories correctly.
Same with other subsystem enabled checks in this method

src/main/java/frc/robot/subsystems/trapElvSubsystem/TrapElvCommandFactory.java
```diff
@@ -142,11 +142,13 @@ public Command positionAMP() {
   public Command scoreAMP() {
     if (subsystem == null) return Commands.none();
     return subsystem
-        .startEnd(
+        .runEnd(
```
azvanderpas, line 145: This should still work with startEnd, right?

azvanderpas (changes requested):

caseyjbrotherton (approved):

PR #215, opened 2024-03-13
Update/west lake

KianChilds:
## Justification
Changes requested during testing at west lake

## Testing Done
Driver practice


src/main/java/frc/robot/OI.java
```diff
@@ -70,7 +70,7 @@ public static void setRumble(double rumbleIntensity) {
   }
 
   public static final class Operator {
-    private static final XboxController controller = new XboxController(operatorJoystickPort);
+    public static final XboxController controller = new XboxController(operatorJoystickPort);
```
caseyjbrotherton, line 73: Talked with @KianChilds about this.  Reason is the getPOV is not implemented in OI.  He will open an issue and reference here for us to fix after next competition

caseyjbrotherton (approved):

Issue #217, opened 2024-03-13
Fix subsystem disabling


azvanderpas:
Disabling some subsystems cause code crashes

PR #220, opened 2024-03-17
Command Factory JUnit tests

azvanderpas:
## Justification

Seen issues crop up from the subsystem disable/enable feature, mixing proxy and non-proxy subsystem Commands, and Commands missing subsystem requirements all together.

#217 

## Implementaion

Added JUnit testing to try to ensure that:
- When the subsystem is null the CommandFactory will not throw NullPointerException
- All Command Factories return a Command that doesn't require a subsystem (before scheduling it)
- All Command Factories require their subsystem when scheduled

Added a getCommands() method to each <Subsystem>CommandFactory which should return an array of all Commands created by the public factory methods with default parameters.

Tried to do this with Reflection to avoid the overhead, but it didn't appear maintainable, especially with generic typed parameters.

## Testing Done

Caught some issues of each type and fixed them in this PR as well


caseyjbrotherton (approved):
Looks good!  Ran locally

PR #214, opened 2024-03-10
Pitch shooting

Josh-Westbrook:
## Justification
 We may have pitch by SA so it'd be good to make sure it'll work

PR #222, opened 2024-03-20
Comp/san antonio

KianChilds:
## Justification
Implementation of the changes requested at comp 

## Testing Done
Ran at comp


src/main/java/frc/robot/RobotContainer.java
```diff
@@ -237,7 +242,14 @@ private void configureBindings() {
     OI.getButton(OI.Operator.retractClimber).toggleOnTrue(climberCommandFactory.climb());
 
     new Trigger(() -> OI.Operator.controller.getPOV() == 0).whileTrue(intakeCommand());
+    new Trigger(() -> OI.Operator.controller.getPOV() == 270)
+        .onTrue(new InstantCommand(() -> robotStateManager.setLongRange()));
```
caseyjbrotherton: Is this the most recent code?  I seem to remember that longrange was unmapped from the commands. @KianChilds

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -237,7 +240,13 @@ private void configureBindings() {
     OI.getButton(OI.Operator.retractClimber).toggleOnTrue(climberCommandFactory.climb());
 
     new Trigger(() -> OI.Operator.controller.getPOV() == 0).whileTrue(intakeCommand());
+    // new Trigger(() -> OI.Operator.controller.getPOV() == 270)
+    //     .onTrue(new InstantCommand(() -> robotStateManager.setLongRange()));
```
caseyjbrotherton, line 244: Let's actually remove this code, and the setShortRange

azvanderpas (changes requested):
Address comments from #214

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -252,24 +261,42 @@ private Command intakeCommand() {
...
   private Command shooterOuttake() {
-
     return Commands.parallel(
             triggerCommandFactory.getEjectCommand(), intakeCommandFactory.reverseIntakeCommand())
         .asProxy();
   }
 
   private void configDriverFeedBack() {
     new Trigger(trapElvCommandFactory.getSourceBreak())
-        .and(OI.getTrigger(OI.Driver.intake))
```
caseyjbrotherton, line 263: We are checking the OI.Driver.intake trigger twice, once to do the intake and another to trigger rumble.

We should see if we can include this in the intake command, or in the command group that is attached to the Intake trigger.

Also, we have a signaling subsystem that has rumble.  Should we use that instead?

src/main/java/frc/robot/RobotContainer.java
```diff
@@ -279,42 +306,71 @@ private Command speakerSource() {
...
   private Command shootSpeaker() {
     return Commands.parallel(
-        Commands.either(
-            triggerCommandFactory.getShootCommand(),
-            triggerCommandFactory
-                .getShootCommand()
-                .onlyIf(() -> shooterCommandFactory.isShooterReady())
-                .asProxy(),
-            OI.getButton(OI.Operator.simple)),
+        Commands.waitSeconds(CommandConstants.WAIT_FOR_TRAPELV)
```
caseyjbrotherton, line 322: What are we waiting for here?  Is this to make sure that isShooterReady is stable?  If so, we should use a debounce instead

src/main/java/frc/robot/subsystems/climberSubsystem/ClimberSubsystem.java
```diff
@@ -53,8 +53,8 @@ private void configMotor(CANSparkMax motor, boolean invert) {
     pidController.setD(ClimberConstants.CURRENT_PID[2]);
     double armPosition;
 
-    armPosition = motor.getAbsoluteEncoder().getPosition();
```
caseyjbrotherton, line 56: So we disabled the absolute encoder during competition? 
Was there a reason why? Are we not expecting to use the absolute encoder in district champs?

src/main/java/frc/robot/RobotContainer.java
KianChilds: we are waiting for the trapelv to move out of the way

src/main/java/frc/robot/subsystems/climberSubsystem/ClimberSubsystem.java
KianChilds: it was occasionally placing the arms at large negative or positive values. Can be reimplemented when needed

src/main/java/frc/robot/RobotContainer.java
KianChilds: to both points: possible, out of scope of this PR

src/main/java/frc/robot/RobotContainer.java
KianChilds: resolved

src/main/java/frc/robot/utilities/AutoManager.java
```diff
@@ -0,0 +1 @@
+
```
caseyjbrotherton, line 1: DElete this file

azvanderpas (changes requested):

src/main/java/frc/robot/subsystems/turretSubsystem/TurretSubsystem.java
```diff
@@ -301,39 +349,36 @@ private static double fineTuneTurretRotation(
...
   public void setTurretPos(double setpoint) {
     if (!Constants.enabledSubsystems.turretRotationEnabled) return;
+    double limitedSetpoint =
+        MathUtil.clamp(
+            setpoint,
+            Math.toRadians(Constants.TurretConstants.TURRET_MIN_ANGLE_DEGREES),
+            Math.toRadians(Constants.TurretConstants.TURRET_MAX_ANGLE_DEGREES));
+    turretGoalPositionEntry.log(Units.radiansToDegrees(setpoint));
+    this.setPositionErrorSupplier(() -> turretPosition - limitedSetpoint);
+  }
```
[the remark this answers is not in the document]
KianChilds: besides controlling the PID through setpoint. Here it is controlled by the error function of the PID. the PIDs goal is always zero making whatever number is returned by the error function the negative of the error processed by the PID.

caseyjbrotherton:
This shouldn't be needed anymore now that newer branches have been merged

PR #230, opened 2024-03-27
Feature/turret result patching

Josh-Westbrook:
## Justification
 To streamline certain parts of the code


src/main/java/frc/robot/subsystems/vision/PhotonSubsystem.java
```diff
@@ -126,51 +122,36 @@ private boolean checkPoseValidity(EstimatedRobotPose pose) {
...
+          lastRecordedTime = turretResult.getTimestampSeconds();
+          return target;
         }
       }
     }
-    if (lastRecordedYawTime != 0
-        && lastRecordedYawTime + Constants.LimelightConstants.APRILTAG_STALE_TIME
-            < Timer.getFPGATimestamp()) {
-      return lastRecordedYaw;
+    return null;
```
azvanderpas, line 136: Should cache a new target if it has one, then should return a cached PhotonTrackedTarget if it is recent enough.

azvanderpas (changes requested):

azvanderpas (approved):

PR #232, opened 2024-03-28
Add servos for climbing

Josh-Westbrook:

### DIFF
--- a/src/main/java/frc/robot/Constants.java
+++ b/src/main/java/frc/robot/Constants.java
@@ -326,10 +326,10 @@ public final class Constants {
   public static class enabledSubsystems {
     public static final boolean intakeEnabled = true;
     public static final boolean drivetrainEnabled = true;
-    public static final boolean visionEnabled = false;
-    public static final boolean usingPhoton = false;
-    public static final boolean elvEnabled = false;
-    public static final boolean signalEnabled = false;
+    public static final boolean visionEnabled = true;
+    public static final boolean usingPhoton = true;
+    public static final boolean elvEnabled = true;
+    public static final boolean signalEnabled = true;
     public static final boolean shooterEnabled = true;
     public static final boolean triggerEnabled = true;
     public static final boolean turretRotationEnabled = true;
--- a/src/main/java/frc/robot/Robot.java
+++ b/src/main/java/frc/robot/Robot.java
@@ -29,7 +29,7 @@ import org.littletonrobotics.junction.wpilog.WPILOGWriter;
  * project.
  */
 public class Robot extends LoggedRobot {
-  public static final boolean isCompetition = false;
+  public static final boolean isCompetition = true;
 
   private Command m_autonomousCommand;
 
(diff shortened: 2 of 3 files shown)


src/main/java/frc/robot/Robot.java
```diff
@@ -29,7 +29,7 @@
  * project.
  */
 public class Robot extends LoggedRobot {
-  public static final boolean isCompetition = true;
+  public static final boolean isCompetition = false;
```
lohengrin332, line 32: This should not be committed.

src/main/java/frc/robot/subsystems/climberSubsystem/ClimberSubsystem.java
```diff
@@ -132,10 +121,59 @@ public CurrentVelocity getVelocity() {
...
+    return state;
+  }
+
+  protected void setState(ClimberState pos) {
+    this.state = pos;
+    updateMotors();
   }
 
+  protected void advanceState() {
+    switch (state) {
```
lohengrin332, line 134: [Java Switch statement documentation]https://www.w3schools.com/java/java_switch.asp).

A `switch` statement with no `break` statements executes each of the items in succession.

Also: you should create a `default` option in case this gets called with `state == CLIP`. It can be a noop or just a log statement, but it should exist.

src/main/java/frc/robot/subsystems/climberSubsystem/ClimberSubsystem.java
```diff
@@ -132,10 +121,59 @@ public CurrentVelocity getVelocity() {
...
+      case MAXIMUM:
+        setState(ClimberState.CLIP);
+      case CLIP:
+        setState(ClimberState.CLIMB);
+    }
+  }
+
+  private void updateMotors() {
+    switch (state) {
```
lohengrin332, line 143: You need breaks and a default here, too.

src/main/java/frc/robot/subsystems/climberSubsystem/ClimberSubsystem.java
```diff
@@ -165,6 +203,22 @@ public void periodic() {
     rightArmOutputEntry.log(rightArmMotor.get());
     leftArmOutputEntry.log(leftArmMotor.get());
     if (this.getCurrentCommand() != null) currentCommand.log(this.getCurrentCommand().getName());
+
+    if (timer.hasElapsed(nextActionTime)) {
+      switch (state) {
```
lohengrin332, line 208: And here.

src/main/java/frc/robot/Constants.java
```diff
@@ -314,9 +326,9 @@ public static class FieldConstants {
   public static class enabledSubsystems {
     public static final boolean intakeEnabled = true;
     public static final boolean drivetrainEnabled = true;
-    public static final boolean visionEnabled = true;
-    public static final boolean usingPhoton = true;
-    public static final boolean elvEnabled = true;
+    public static final boolean visionEnabled = false;
```
lohengrin332, line 329: These should not be committed as `false`.

PR #235, opened 2024-03-29
Remove ground beam break + Rename sourceBreak

ModBoyex:
## Justification
There were plans to have two beam breaks at some point, but this never happened, and this code has been left unused for the whole time. It's time to move on and get rid of it.

## Testing Done
It builds without the code, so we should be good to go.

### DIFF
--- a/src/main/java/frc/robot/subsystems/trapElvSubsystem/TrapElvSubsystem.java
+++ b/src/main/java/frc/robot/subsystems/trapElvSubsystem/TrapElvSubsystem.java
@@ -57,8 +57,7 @@ public class TrapElvSubsystem extends SubsystemBase {
   private double scoringMotorOffset;
 
   // Beam Breaks
-  private final TOFSensorSimple sourceBreak;
-  private final DigitalInput groundBreak;
+  private final TOFSensorSimple wristBeamBreak;
 
   // Limit Switches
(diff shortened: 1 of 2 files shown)


src/main/java/frc/robot/subsystems/trapElvSubsystem/TrapElvSubsystem.java
```diff
@@ -262,31 +258,22 @@ public TrapElvSubsystem() {
         new DebugEntry<Double>(0.0, "Wrist Motor Output", this).withPosition(2, 1).withSize(2, 1);
     FFOutput = new DebugEntry<Double>(0.0, "FF Output", this).withPosition(5, 1).withSize(2, 1);
     sourceLog =
-        new DebugEntry<Boolean>(sourceBreak.isBeamBroke(), "Source BB", this).withPosition(0, 2);
-    groundLog = new DebugEntry<Boolean>(groundBreak.get(), "Ground BB", this).withPosition(1, 2);
+        new DebugEntry<Boolean>(wristBeamBreak.isBeamBroke(), "Source BB", this).withPosition(0, 2);
```
caseyjbrotherton, line 261: Do we want to change this to Wrist BB?

caseyjbrotherton (approved):

PR #236, opened 2024-03-29
Tunnable number subscribers

ModBoyex:
## Justification
Tunable numbers have been taking up too much CPU, so we are making use of Subscribers rather than periodically pulling from NT.

## Testing Done
Needs to be testing on real robot, but has been simulated, and seems to be working.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,12 +30,12 @@ deploy {
 
                 frcJava(getArtifactTypeClass('FRCJavaArtifact')) {
                     // Unomment to enable VisualVM connection
-                    jvmArgs.add("-Dcom.sun.management.jmxremote=true")
-                    jvmArgs.add("-Dcom.sun.management.jmxremote.port=1198")
-                    jvmArgs.add("-Dcom.sun.management.jmxremote.local.only=false")
-                    jvmArgs.add("-Dcom.sun.management.jmxremote.ssl=false")
-                    jvmArgs.add("-Dcom.sun.management.jmxremote.authenticate=false")
-                    jvmArgs.add("-Djava.rmi.server.hostname=10.63.77.2") // Replace XX.XX with team number
+                    // jvmArgs.add("-Dcom.sun.management.jmxremote=true")
+                    // jvmArgs.add("-Dcom.sun.management.jmxremote.port=1198")
+                    // jvmArgs.add("-Dcom.sun.management.jmxremote.local.only=false")
+                    // jvmArgs.add("-Dcom.sun.management.jmxremote.ssl=false")
+                    // jvmArgs.add("-Dcom.sun.management.jmxremote.authenticate=false")
+                    // jvmArgs.add("-Djava.rmi.server.hostname=10.63.77.2") // Replace XX.XX with team number
                 }
 
                 // Static files artifact
--- a/build.gradle
+++ b/build.gradle
@@ -30,12 +30,12 @@ deploy {
 
                 frcJava(getArtifactTypeClass('FRCJavaArtifact')) {
                     // Unomment to enable VisualVM connection
-                    // jvmArgs.add("-Dcom.sun.management.jmxremote=true")
-                    // jvmArgs.add("-Dcom.sun.management.jmxremote.port=1198")
-                    // jvmArgs.add("-Dcom.sun.management.jmxremote.local.only=false")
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -30,12 +30,12 @@ deploy {
 
                 frcJava(getArtifactTypeClass('FRCJavaArtifact')) {
                     // Unomment to enable VisualVM connection
-                    // jvmArgs.add("-Dcom.sun.management.jmxremote=true")
-                    // jvmArgs.add("-Dcom.sun.management.jmxremote.port=1198")
-                    // jvmArgs.add("-Dcom.sun.management.jmxremote.local.only=false")
-                    // jvmArgs.add("-Dcom.sun.management.jmxremote.ssl=false")
-                    // jvmArgs.add("-Dcom.sun.management.jmxremote.authenticate=false")
-                    // jvmArgs.add("-Djava.rmi.server.hostname=10.63.77.2") // Replace XX.XX with team number
+                    jvmArgs.add("-Dcom.sun.management.jmxremote=true")
```
caseyjbrotherton: comment these back for main

src/main/java/frc/robot/Robot.java
```diff
@@ -29,7 +29,7 @@
  * project.
  */
 public class Robot extends LoggedRobot {
-  public static final boolean isCompetition = true;
+  public static final boolean isCompetition = false;
```
caseyjbrotherton: set to isCompetition for main

src/main/java/frc/robot/utilities/TunableNumber.java
```diff
@@ -25,21 +36,22 @@ public TunableNumber(
...
+      doubleSub = doubleTopic.subscribe(defaultValue, PubSubOption.pollStorage(0), PubSubOption.periodic(1));
     }
   }
 
   public void periodic() {
     if (!Robot.isCompetition) {
-      consumer.accept(numberEntry.getDouble(value));
-      value = numberEntry.getDouble(value);
+      consumer.accept(doubleSub.get(defaultValue));
+      value = doubleSub.get(defaultValue);
```
azvanderpas, line 51:
```suggestion
      value = doubleSub.get(defaultValue);
      consumer.accept(value);
```
Even though we've made it more network efficient, its still better to only request it when we need it

src/main/java/frc/robot/utilities/TunableNumber.java
```diff
@@ -25,21 +36,22 @@ public TunableNumber(
     this.value = defaultValue;
     this.consumer = consumer;
     if (!Robot.isCompetition) {
+      inst = NetworkTableInstance.getDefault();
       numberEntry = tuningTab.add(name, defaultValue).getEntry();
+      numberEntry.close();
+      
+      doubleTopic = inst.getDoubleTopic(name);
```
azvanderpas: Since we get the topic from elsewhere, the GenericEntry isn't needed anymore.

```suggestion
      inst = NetworkTableInstance.getDefault();
      tuningTab.add(name, defaultValue);
      
      doubleTopic = inst.getDoubleTopic(name);
```

src/main/java/frc/robot/utilities/TunableNumber.java
```diff
@@ -10,11 +16,16 @@
 import java.util.function.DoubleSupplier;
 
 public class TunableNumber extends SubsystemBase implements DoubleSupplier {
+  private NetworkTableInstance inst;
   private static ShuffleboardTab tuningTab;
   private GenericEntry numberEntry;
   private double value;
+  private double defaultValue;
```
azvanderpas, line 21: final?

src/main/java/frc/robot/utilities/TunableNumber.java
```diff
@@ -10,11 +16,16 @@
 import java.util.function.DoubleSupplier;
 
 public class TunableNumber extends SubsystemBase implements DoubleSupplier {
+  private NetworkTableInstance inst;
```
azvanderpas, line 17: Could also be static

azvanderpas (changes requested):

src/main/java/frc/robot/utilities/TunableNumber.java
ModBoyex: it gets made if the var is not inisialized when comp is on

azvanderpas (approved):

PR #238, opened 2024-03-30
Feature/modes

KianChilds:


src/main/java/frc/robot/subsystems/turretSubsystem/TurretCommandFactory.java
```diff
@@ -137,7 +138,34 @@ public Command zeroTurretCommand() {
...
+          switch (RSM.getShooterMode()) {
+            case LOB:
+              return lobShot();
+            case LONG_RANGE:
+              return longRangeShot();
+            case SHORT_RANGE:
+              return shortRangeShot();
+            default:
+              Runnable errMsg = () -> DriverStation.reportError("Unknown Shooter Mode!", true);
+              return new InstantCommand(errMsg);
```
lohengrin332, line 153: What happens when you run this error command? I think it would be preferable to return a valid default, either `LONG_RANGE` or `SHORT_RANGE`. Could we have a single command which runs `reportError`, and _then_ runs the default shooter aim mode command?

lohengrin332 (changes requested):
The spotless checks failed. Please run `Build Robot Code` and commit the changes to correct this.

I also have one question.

src/main/java/frc/robot/subsystems/turretSubsystem/TurretCommandFactory.java
lohengrin332: Update: after talking with Casey about this, it makes more sense to report the error immediately, then return a good default value (short range). I'll make these changes.

src/main/java/frc/robot/stateManagement/RobotStateManager.java
```diff
@@ -139,4 +124,21 @@ public DoubleSupplier getSpeakerAngle() {
...
+    return shooterMode;
+  }
+
+  public void setShooterMode(ShooterMode shooterMode) {
+    this.shooterMode = shooterMode;
+  }
+
+  public Translation2d getLobPosition() {
+    // TODO Auto-generated method stub
+    throw new UnsupportedOperationException("Unimplemented method 'getLobPosition'");
```
lohengrin332, line 138: I think this needs to be implemented before we can merge this code. I thought we talked about it being the same angle as short range mode, which would make it easy to implement, right?

lohengrin332 (approved):

PR #245, opened 2024-03-31
Pidgeon pose

ModBoyex:
## Justification
Our odometry is not accurate when we are being heavily defended, and this causes our turret to be off because it does not know the correct location for the robot. The Pigeon however has an accelerometer along with a gyro, this could in theory give us a position that would account for being deffended.

## Implementaion
All this branch does is log the theoretical pose that the Pigeon would give us if we were to use it, just to see how noisy/usable it might be.


src/main/java/frc/robot/subsystems/swerveSubsystem/SwerveSubsystem.java
```diff
@@ -194,8 +198,23 @@ public double getMagnitude() {
...
 
+  private double retX[] = new double[2];
+  private double retY[] = new double[2];
+  private Pose2d pidgeonPose = new Pose2d();
+  private Field2d pdigeonField = new Field2d();
+
   @Override
   public void periodic() {
+    retX =
+        HowdyMath.getPidgeonTranslation(
```
caseyjbrotherton, line 209: Mis-spell

PR #256, opened 2024-08-24
Proof of Concept: Submodule as Source Library

azvanderpas:
## Justification

Would like to have a more convenient way to reuse and build upon code year over year. 

## Implementation

By using a new repo as a [submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules) that can be installed in any season repo, common code can exists in one location while being used in any year's codebase. Submodules in a repo are tracked by the submodule's commit hash, so future changes won't break previous year's builds. 

Using submodules does add a step to the 'git clone' process. After cloning, 'git submodule update --init' needs to be run to clone the submodule files.

Submodules must be confined to one directory, but the Java project file locations we'd want to include are at a few different locations. To solve this, I added task in build.gradle that will automatically link the files into the correct locations when a build is run. This is done using git-bash on Windows using 'ln -P'. Open to suggestions for alternatives, but it seemed to work fairly smoothly. Any changes to the files at the linked locations is reflected within the submodules location (and vice versa).

Check out [frcsubmodule](https://github.com/azvanderpas/frcsubmodule) for the submodule repo organization and Makefile for linking the files. This is just a proof of concept (i.e. don't merge this), so I used my personal github to make a submodule as a test.

## Testing Done

Linking and building tested to work in Windows environment (assuming git-bash.exe exists) and passes CI using Ubuntu Linux system.


eskiles (approved):
The changes to the git and gradle files seem easy enough to repeat for other teams.

I really like that there is a single repository to maintain truly common utility code that can be applied to many robots through the season. This is better than trying to get new students to copy and paste from other repos.

I am trying out the fluent builder pattern for PID and Feedforward data objects with my students. I'll let you all know how that goes this year.

